PR #334, opened 2015-01-07
Add exit codes to rake task.

wli:
I couldn't figure out how to mock a file correctly, so I don't have a positive test case. I did test this locally and it returns things as expected:

````
wli@oort:~/projects/test$ rake scss_lint
app/assets/stylesheets/test.scss:20 [W] SingleLinePerSelector: Each selector in a comma sequence should be on its own line
scss-lint found lints
wli@oort:~/projects/test$ echo $?
1
````

### DIFF
--- a/lib/scss_lint/rake_task.rb
+++ b/lib/scss_lint/rake_task.rb
@@ -29,12 +29,15 @@ module SCSSLint
       require 'scss_lint'
       require 'scss_lint/cli'
 
-      CLI.new.run([])
+      exit CLI.new.run([])
     rescue SystemExit => ex
-      if ex.status == CLI::EXIT_CODES[:data]
-        abort('scss-lint found lints')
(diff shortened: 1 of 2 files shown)


sds:
Hey @wli, thanks for the pull request.

While my first reaction to this is that rake tasks should not return status codes, I appreciate the fact that this isn't a hard rule outlined anywhere. Bear in mind this does prevent `scss-lint` from being used as a dependency for another rake task.

However, given I'm pretty sure the major use case for rake integration is to simply run `rake scss_lint` to lint, I'm fine including this for now. If this bites anyone else, feel free to open an issue and we'll revisit.

Merged in 6aa9703 with some adjustments in bd0ffa9.

wli:
Thanks for fixing this up @sds! Looking forward to the release.

PR #340, opened 2015-01-15
Add SelectorFormat convention for CSS Wizardry

bookwyrm:
CSS Wizardy has a variation on the BEM SelectorFormat which is different enough from the BEM standard to warrant a different validation regex.

http://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -105,7 +105,7 @@ linters:
 
   SelectorFormat:
     enabled: true
-    convention: hyphenated_lowercase # or 'BEM', or 'snake_case', or 'camel_case', or a regex pattern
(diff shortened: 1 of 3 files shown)


Anahkiasen:
Isn't that already `BEM_hyphenated` ?

bookwyrm:
The main differences are in the use of double hyphen for the modifier (`.block__element--modifier`) and the ability to put an element after a modifier (`.block--modifier__element`).

bookwyrm:
Oh, I hadn't seen `hyphenated_BEM`...

bookwyrm:
Ok, yes, `hyphenated_BEM` works for my purposes. I'll submit a new pull request with tests and to expose the convention in the default config.

PR #342, opened 2015-01-15
Tests and exposed config option for hyphenated_BEM

bookwyrm:
Unless you look at the source, you don't realize that there is a SelectorFormat convention named hyphenated_BEM.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -105,7 +105,7 @@ linters:
 
   SelectorFormat:
     enabled: true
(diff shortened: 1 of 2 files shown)


lencioni:
:+1: I'm actually surprised we didn't have tests for `hyphenated_BEM`.

sds:
Thanks for the pull request, @bookwyrm!

We really appreciate you adding tests for this feature!

Merged in c955f70cd12cf64cae77dabe37083f3dd0e783c5.

PR #348, opened 2015-01-22
Add linter for important rule

mikeholm:


sds:
Thanks for the pull request, @mikeholm! 

I've merged this in 10f9c68 and enabled it by default in a6abf87. I think this is valuable to enable by default given how often `!important` is (still) misused, and the fact that you can use a control comment to disable the lint for special cases.

PR #349, opened 2015-01-22
Add linter for variables in property values

mikeholm:


lib/scss_lint/linter/use_variables_for_properties.rb
```diff
@@ -0,0 +1,18 @@
+module SCSSLint
+  class Linter::UseVariablesForProperties < Linter
+    include LinterRegistry
+
+    def visit_root(_node)
+      @included_properties = Set.new(config['included_properties'])
```
sds: Can just call this `properties`. Since there is no concept of `excluded_properties`, the `included_` qualifier seems unnecessary.

lib/scss_lint/linter/use_variables_for_properties.rb
```diff
@@ -0,0 +1,18 @@
...
+      yield
```
sds: Since this linter only serves a purpose if there are properties defined, you can short circuit running anything at all by writing:

```ruby
yield if @properties.any?
```

lib/scss_lint/linter/use_variables_for_properties.rb
```diff
@@ -0,0 +1,18 @@
...
+    end
+
+    def visit_prop(node) 
+      return if node.name.count > 1 || !node.name.first.is_a?(String)
+      return unless @included_properties.include? node.name.first
+      return if node.children.first.is_a?(Sass::Script::Tree::Variable)
```
sds: What about properties that are the result of a function call? I'm open to thoughts on this, but I can see the value of making this strict about requiring variables.

sds:
Thanks for the pull request, @mikeholm.

Happy to merge once the Travis build is fixed and my comments are addressed. Overall looks great!

lib/scss_lint/linter/use_variables_for_properties.rb
mikeholm: That's a great point. How do you feel about leaving it to variables for now and then figuring out the function thing at some other time? Would that cause trouble by possibly breaking your API?

mikeholm:
Alright, Travis is happy and I think I've addressed your concerns. Take a look?

Thanks for this awesome tool and taking our PRs!

sds:
Thanks for addressing those comments, @mikeholm.

Merged in 53d4533 with some adjustments in 8c85f9f. Main difference is renaming the linter to match the conventions laid out in other linters.

Thanks for your contribution to `scss-lint`!

PR #354, opened 2015-01-30
Wrap at-declaratives in backticks

teoljungberg:
We, at thoughtbot, use `scss-lint` to check for style violations in
[Hound][hound]. Hound will comment on style violations on an open PR,
and leave comments on the violating lines.

So to avoid pinging GitHub users in the comment, I suggest wrapping the
declaratives in backticks.

### DIFF
--- a/lib/scss_lint/linter/declaration_order.rb
+++ b/lib/scss_lint/linter/declaration_order.rb
@@ -15,8 +15,8 @@ module SCSSLint
 
     MESSAGE =
       'Rule sets should be ordered as follows: '\
-      '@extends, @includes without @content, ' \
-      'properties, @includes with @content, ' \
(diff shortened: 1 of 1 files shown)


trotzig:
Hi @teoljungberg! I'm glad to hear Thoughtbot is using scss-lint! (I've heard that you're also using React, which is even more awesome, but that's unrelated). 

Wrapping these strings in backticks seems like a good idea. I'm just running the commit through an internal code review tool, but expect it to be merged soon.

trotzig:
Merged in https://github.com/causes/scss-lint/commit/09eee7e42d139602d6c94127b8a3a20c27a17e1a. Thanks for the contribution!

teoljungberg:
Thank you @trotzig, glad I could help out

PR #371, opened 2015-02-20
Documentation and specs for NameFormat with regexp convention

Dorian:
It was not obvious what the format of the regexp should be


lencioni:
Thanks for the contribution! Merged in 52e69ac7c7766729cff4806502777df026b2e871

Dorian:
That was fast! Thanks, BTW I love scss-lint

PR #372, opened 2015-02-20
Add a new CheckStyle formatter

Sweetchuck:
I have not found any Jenkins plugin which is suitable to process an XML which is generated by the `--format=XML` option.

This new formatter/reporter almost the same as the XML formatter only difference is that this produce a standard Checkstyle format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,6 +283,27 @@ scss-lint --format=XML [scss-files...]
 </lint>
 ```
 
+### CheckStyle
+
+Outputs a CheckStyle XML with `<checkstyle>`, `<file>`, and `<error>` tags. Suitable for
+consumption by tools like [Jenkins](http://jenkins-ci.org/) with [CheckStyle plugin](https://wiki.jenkins-ci.org/display/JENKINS/Checkstyle+Plugin).
+
+```bash
+scss-lint --format=CheckStyle [scss-files...]
+```
+
+```xml
+<?xml version="1.0" encoding="utf-8"?>
+<checkstyle version="1.5.6">
+  <file name="test.css">
+    <error line="2" severity="warning" message="Prefer single quoted strings" />
+    <error line="2" severity="warning" message="Line should be indented 0 spaces, but was indented 1 spaces" />
+    <error line="5" severity="warning" message="Prefer single quoted strings" />
+    <error line="6" severity="warning" message="URLs should be enclosed in quotes" />
+  </file>
+</checkstyle>
+```
+
 ## Exit Status Codes
 
 `scss-lint` tries to use
--- a/lib/scss_lint/reporter/checkstyle_reporter.rb
+++ b/lib/scss_lint/reporter/checkstyle_reporter.rb
@@ -1,0 +1,33 @@
+module SCSSLint
+  # Reports lints in a CheckStyle format.
+  class Reporter::CheckStyleReporter < Reporter
+    def report_lints
+      output = '<?xml version="1.0" encoding="utf-8"?>'
+
+      output << '<checkstyle version="1.5.6">'
+      lints.group_by(&:filename).each do |filename, file_lints|
+        output << "<file name=#{filename.encode(xml: :attr)}>"
+
+        file_lints.each do |lint|
+          output << issue_tag(lint)
+        end
+
+        output << '</file>'
+      end
+      output << '</checkstyle>'
+
+      output
+    end
+
+  private
+
+    def issue_tag(lint)
+      "<error source=\"#{lint.linter.name if lint.linter}\" " \
+             "line=\"#{lint.location.line}\" " \
+             "column=\"#{lint.location.column}\" " \
+             "length=\"#{lint.location.length}\" " \
+             "severity=\"#{lint.severity}\" " \
+             "message=#{lint.description.encode(xml: :attr)} />"
+    end
+  end
+end
--- a/spec/scss_lint/reporter/checkstyle_reporter_spec.rb
+++ b/spec/scss_lint/reporter/checkstyle_reporter_spec.rb
@@ -1,0 +1,103 @@
+require 'spec_helper'
+
+describe SCSSLint::Reporter::CheckStyleReporter do
+  subject { SCSSLint::Reporter::CheckStyleReporter.new(lints) }
+
+  describe '#report_lints' do
+    let(:xml) { Nokogiri::XML(subject.report_lints) }
+
+    shared_examples_for 'XML document' do
+      it 'has an encoding of UTF-8' do
+        xml.encoding.should == 'utf-8'
+      end
+
+      it 'has an XML version of 1.0' do
+        xml.version.should == '1.0'
+      end
+
+      it 'contains a <checkstyle> root element' do
+        xml.root.name.should == 'checkstyle'
+      end
+    end
+
+    context 'when there are no lints' do
+      let(:lints) { [] }
+
+      it_should_behave_like 'XML document'
+    end
+
+    context 'when there are lints' do
+      let(:filenames)    { ['f1.scss', 'f2.scss', 'f1.scss'] }
+      # Include invalid XML characters in the third description to validate
+      # that escaping happens for preventing broken XML output
+      let(:descriptions) { ['lint 1', 'lint 2', 'lint 3 " \' < & >'] }
+      let(:severities)   { [:warning] * 3 }
+
+      let(:locations)    do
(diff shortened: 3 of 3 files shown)


README.md
```diff
@@ -283,6 +283,27 @@ scss-lint --format=XML [scss-files...]
...
+Outputs a CheckStyle XML with `<checkstyle>`, `<file>`, and `<error>` tags. Suitable for
+consumption by tools like [Jenkins](http://jenkins-ci.org/) with [CheckStyle plugin](https://wiki.jenkins-ci.org/display/JENKINS/Checkstyle+Plugin).
+
+```bash
+scss-lint --format=CheckStyle [scss-files...]
+```
+
+```xml
+<?xml version="1.0" encoding="utf-8"?>
+<checkstyle version="1.5.6">
```
sds: Including the version number here seems brittle. Is this absolutely necessary (i.e. does the CheckStyle plugin require this attribute be present)?

Having `scss-lint` need to update this formatter's output each time the `CheckStyle` plugin is updated isn't ideal.

Sweetchuck: > Does the CheckStyle plugin require this attribute be present?

I do not know, but this is how the PHP_CodeSniffer works. It also puts this version number into the generated XML. Maybe this indicates that the XML which version of the Checkstyle definition compatible with. 
The version number has no any direct relation to the Jenkins Checkstyle plugin.

sds: The version number you've specified seems really out of date. The current version of the CheckStyle plugin is 3.41.

I guess my issue with this is that this plugin is extremely specific to a particular testing platform (Jenkins) and isn't a generic output format (e.g. JSON).

I would rather this be pulled into its own separate gem and loaded via the `--require` and `--format` flags. I don't think this formatter should be shipped with `scss-lint`.

sds:
Thanks for the pull request, @Sweetchuck. We really appreciate having developers create integrations for `scss-lint` with other toolsets.

In this case, I think this formatter is specific enough to warrant its own gem, which can then be loaded with the `--require` and `--format` flags. This relieves us of the burden of maintaining it, but allows others to use it.

Sweetchuck:
Sorry, I am not agree with you.
I think the current `--format=XML` is useless because there is no such a tool which is able to use it.
You mention that in the `readme.md` about the XML output
 "Suitable for consumption by tools like Jenkins." https://github.com/causes/scss-lint#xml 
But I have not found any plugin which can do anything with that XML.

The Checkstyle is a standard format and widely used, just with Jenkins more than 12.000 CI server instance use it. https://wiki.jenkins-ci.org/display/JENKINS/Checkstyle+Plugin

http://ant.apache.org/ and https://www.phing.info are also able to mark a build as _unstable_ based on a checkstyle.xml file.

I really do not understand why you don't want to support the checkstyle format.

sds:
I apologize--that documentation is clearly incorrect.

When the XML formatter was first added (it was actually the first pull request to `scss-lint`; see PR #1) we merged it without verifying the author's claim that it was intended to be consumed by Jenkins. Thus we included it in the documentation without confirming ourselves that it was "useful". It has stuck around ever since.

I've removed the confusing documentation in 8ea0292f234610079e0a00a0f036089e19ed5a6d. You'll have to talk to @volpino to understand how he is using it with Jenkins.

Now let me explain why I don't want to merge this into the main `scss-lint` project.

By merging this, we are setting a precedent that `scss-lint` directly supports third-party tools (like the CheckStyle Jenkins plugin). This is _not_ a claim we as a maintainer of the gem want to make. As we don't use the CheckStyle plugin internally, we can't promise to ensure it's well-tested and functional.

Thus the approach we suggest is that some other party (e.g. yourself) take the file `lib/scss_lint/reporter/checkstyle_reporter.rb`, place it into a separate gem, and then distribute that gem. We will happily include a link to it in our documentation. Anyone can then use the formatter by passing the `--require=scss-lint-checkstyle-reporter --format=Checkstyle` flags to `scss-lint`. This provides a clean way for reuse while relieving us the burden of maintenance. This is a style that many open source projects use (see the [Atom Linter](https://atom.io/packages/linter) project as an example).

These are the sorts of decisions any large open source project needs to make. I'm sorry if this seems unfair, but I believe it's in the best interest of the project.

Sweetchuck:
Atom linter is bad example because it aggregates other linters.
 
http://eslint.org/docs/command-line-interface/#-f,---format

http://jshint.com/docs/cli --reporter option

mprins:
even though this was closed I would like to cast a +1 vote for merging this, and optionally dropping/replacing the existing xml formatter (which is fairly useless IMO)

The checkstyle output is well supported in many enterprise build and code quality reporting tools, Jenkins being just a single example, a longer list is available at http://checkstyle.sourceforge.net/index.html#Related_Tools

sds:
@mprins I'm down for dropping the XML formatter. It's an old relic that in light of this new information isn't being used. (@volpino: you're the only known entity who might be using it; if so, speak up now)

I'm trying to understand why this is being met with such resistance. Let's be clear: **I want `scss-lint` to be easy to integrate with popular tools to make developer's lives easier.** My opposition to this pull request is that it:

- Puts the burden of maintaining a third-party integration on the `scss-lint` project (an integration that none of the maintainers use)
- Couples `scss-lint` code to a third-party plugin (if that plugin introduces a breaking change, the integration in `scss-lint` breaks)
- Ties any bug fixes to the `scss-lint` release cycle. For one third-party integration this is not a big deal, but down the line if we had many integrations there could be breaking changes happening all the time

The argument that this is supported by other tools and thus `scss-lint` should also support it is an altered form of the [bandwagon fallacy](https://en.wikipedia.org/wiki/Argumentum_ad_populum).

I understand that having this integration be a first-class citizen of `scss-lint` is convenient for developers, but what I am suggesting is not outrageous. For other developers who would want to use the checkstyle integration, it changes the list of steps from:

```
scss-lint --format=CheckStyle
```

...to:

```
gem install scss-lint-reporter-checkstyle # One time operation
scss-lint --require=scss-lint-reporter-checkstyle --format=CheckStyle 
```

You write your Jenkins job once and never think about this again. I don't see how this is objectionable.

If the concern is that you then have to maintain this integration, then **now you understand where I'm coming from!** :wink: 

I don't think third-party integrations belong as first-class citizens in tools _if_ there is a convenient way to support them as plugins. It introduces unnecessary coupling which is poor software design.

Here's my offer: I am happy to write this gem for you and transfer ownership (this will take 10 minutes), and add documentation in the README with instructions on how to use the gem. Is that reasonable?

mprins:
:blush:  yeah, convenience for me is my main drive  :blush: 

definitely more than reasonable!

I have no objections to taking ownership and would probably pulled something like this into the maven plugin* I'm working on atm anyways.

*see: https://github.com/GeoDienstenCentrum/sass-maven-plugin/pull/3

volpino:
@sds hi! unfortunately i don't work anymore in the company where they were/are using scss-lint with Jenkins. Maybe @nostalgiaz know more than me on the matter.

Sweetchuck:
How often changing the format of the jUnit or Checkstyle XML without backward compatibility?

If you are going to drop the XML output you should drop the JSON output as well, because it also suppose a third-part integration. (I do not know any tool which is able to work with such an output)

I think the "scss-lint" should support the widely used output formats __out of box__, this is what the other linter tools do. And beside those formats one can use a _customformat.js_

http://eslint.org/docs/command-line-interface/#-f,---format

`-f, --format`
This option specifies the output format for the console. Possible formats are "stylish" (the default), "compact", "checkstyle", "jslint-xml", "junit" and "tap".
Example:
`eslint -f compact file.js`


You can also use a custom formatter from the command line by specifying a path to the custom formatter file.
Example:
`eslint -f ./customformat.js file.js`

Sweetchuck:
@sds Before I publish a new gem on rubygems.org I would like to ask you to give a code review to this repository. Thank you.

https://github.com/Sweetchuck/scss_lint_reporter_checkstyle

sds:
Thanks for setting up the gem, @Sweetchuck. 

I've left some comments in https://github.com/Sweetchuck/scss_lint_reporter_checkstyle/commit/9c6f20efc85a6b53701920089df25db2bf921ba2. I don't think there are any blockers, so feel free to fire when ready!

Sweetchuck:
Thank you again.
https://rubygems.org/gems/scss_lint_reporter_checkstyle

sds:
Added documentation for how to use plugin in 0c2dbfc.

PR #385, opened 2015-03-06
Add animation* properties to concentric.txt

loadedsith:
PropertySortOrder:
      order: concentric 

First suggests 

    Properties should be ordered animation-name, animation-fill-mode

So one updates the order, it then suggests

    Properties should be ordered animation-fill-mode, animation-name

This is an infinite loop (assuming the human doesn't notice, lol).

Upon inspection, the animation rules are missing from the contentric.txt setting, this PR adds them.

Using the `animation` shorthand as a key, this is the order I came up with:

animation
animation-name
animation-duration
animation-timing-function
animation-delay
animation-iteration-count
animation-direction
animation-fill-mode
animation-play-state

### DIFF
--- a/data/property-sort-orders/concentric.txt
+++ b/data/property-sort-orders/concentric.txt
@@ -122,3 +122,12 @@ font-style
 
 content
 quotes
+
+animation
+animation-name
+animation-fill-mode
+animation-delay
+animation-duration
+animation-iteration-count
+animation-play-state
+animation-timing-function


lencioni:
Thanks for the contribution! This has been merged in fd1cc640d5cc517eb2bc79dab19bd34fb1d3aaf1

I'm not very familiar with the concentric ordering, but might it make sense to put these close to `transition`?

lencioni:
Perhaps @Anahkiasen has an opinion?

lencioni:
Addressed by #386

PR #386, opened 2015-03-06
Move animation rules near to transition rules

loadedsith:
As per suggestion, animations do belong near transition rules


lencioni:
Thanks! Merged in 544d39ad33581afd4e5852d41d5eaa8a1b5e4849

PR #396, opened 2015-03-19
Fix allow_non_nested_indentation.

kristijanhusak:
This is the fix for https://github.com/causes/scss-lint/issues/374 .

I tested it on some basic examples, and it works fine, i'm just not sure if it breaks something else.

Tests are passing, so i believe it's ok.

### DIFF
--- a/lib/scss_lint/linter/indentation.rb
+++ b/lib/scss_lint/linter/indentation.rb
@@ -49,9 +49,7 @@ module SCSSLint
         return true
       end
 
-      unless allow_arbitrary_indent?(node) ||
-          actual_indent.length == @indent ||
-          actual_indent.length % @indent_width == 0
+      unless allow_arbitrary_indent?(node, actual_indent.length) || actual_indent.length == @indent
         add_lint(node.line,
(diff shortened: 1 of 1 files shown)


sds:
Thanks for the pull request, @kristijanhusak.

Unfortunately, the problem is more complicated than you think. Your code fails on the following input:

```scss
.one {
  color: #000;
}
  .two {
      margin: 0;  // This is indented too much, but doesn't report a lint
  }
```

In essence, you need to compare the indentation of a node to its parent's indentation to truly know if it has the correct indentation, since the parent could be arbitrarily indented.

I've implemented a solution in 53494a8, since there seems to be enough people wanting a fix for this issue.

kristijanhusak:
@sds Yeah, you're right, I was just relying on tests. I'm not very familiar with Sass API, and this is my first time I tried to do anything in ruby :)

Thanks for the fix. Will you tag a release so we can update it?

sds:
Version [0.35.0](https://github.com/causes/scss-lint/releases/tag/v0.35.0) has been released.

PR #407, opened 2015-03-31
Add PropertyUnits rule

imcnally:
Hi all,

I originally added a NoPixels rule, but after @sds 's suggestion, I turned it into `PropertyUnits`. It allows you set globally allowed units, and override them on a per-property basis.

This is designed to enforce the use, or lack of use, of certain units, i.e., disallowing pixels over REMs, or not allowed unit-based line heights.

- Ian


Anahkiasen:
:+1: for this, absolutely agree. I would maybe add a rule to disable it on certain properties (like the variables linters), I know our designers use em/rem/etc on most things but pixels on certain things that do require them so that'd allow us to use the linter efficiently.

imcnally:
great idea @Anahkiasen. Added it in  1eb3cb8.

sds:
Hey @imcnally, thanks for the pull request.

I think taking this a step further would actually result in this idea being more useful in general.

What if we were to rename this linter to `PropertyUnits`, and expose configuration options to specify which units are allowed not just globally, but also on a per-property basis.

For example, perhaps it makes sense to always use pixels for borders, but `em`/`rem` are best for fonts. Each team will have different preferences on which units they want to enforce for different property types.

Thus a typical configuration for this linter would look something like:

```yaml
linters:
  Units:
    global: ['em', 'ex', '%', 'px', 'ch', 'cm', 'mm', 'in', 'pt', 'pc', 'rem', 'vh', 'vw', 'vmin', 'vmax'] # Allow any units globally
    properties: 
      margin: ['em', 'rem']
      border: ['px']
      line-height: [] # No units allowed
```

By default the linter will allow all units by default, and developers can adjust the `global`ly allowed units or set per-property exceptions as they see fit. The allowed units for a given property is the set intersection of `global` and that property's list of allowed units (if none is defined, use `global`).

What do you think?

nenadjelovac:
@sds so effin awesome! +1

imcnally:
Good idea @sds. I was looking at it from a disallow-these-things perspective, but I like the idea of globally allowed properties. It makes you consider what units you can and/or want to use more. And a property-by-property override is a good compliment to that.

config/default.yml
```diff
@@ -98,6 +98,11 @@ linters:
     include_nested: false
     max_properties: 10
 
+  PropertyUnits:
+    enabled: false
+    global: ['em', 'ex', '%', 'px', 'ch', 'cm', 'mm', 'in', 'pt', 'pc', 'rem', 'vh', 'vw', 'vmin', 'vmax']
+    properties: []
```
sds: This should be an empty hash rather than an empty array.

lib/scss_lint/linter/property_units.rb
```diff
@@ -0,0 +1,59 @@
...
+    include LinterRegistry
+
+    def visit_prop(node)
+      @global_allowed = config['global'].to_set
+      @properties = properties config
+      property = node.value.value.to_s
+      name = node_name node
+      units = property.scan /[a-zA-Z%]+/ix
+
+      return if units.empty?
```
sds: This gives me an idea for adding support for checking if a property where units are required is lacking units, but that might be harder to pull off. (just thinking out loud)

lib/scss_lint/linter/property_units.rb
```diff
@@ -0,0 +1,59 @@
...
+    def node_name node
+      return dash_to_underscore node.name.join
+    end
+
+    def dash_to_underscore str
+      return str.sub('-', '_')
+    end
+
+  end
+end
```
sds: Can you run `bundle exec rubocop` against this repo? We're supposed to have Travis report back offenses but it's currently broken, and there are a few style nits here that would be reported by that.

spec/scss_lint/linter/property_units_spec.rb
```diff
@@ -0,0 +1,168 @@
+require 'spec_helper'
+
+describe SCSSLint::Linter::PropertyUnits do
+
+  context 'when global units are set and local are not set' do
+    let(:linter_config) {{
+      'global' => ['rem'],
+      'properties' => []
```
sds: Should be an empty hash.

lib/scss_lint/linter/property_units.rb
```diff
@@ -0,0 +1,59 @@
...
+      return any_units_not_allowed? units, @properties[property]
+    end
+
+    def any_units_not_allowed? units, allowed
+      return units.select {|unit| !allowed.include?(unit)}
+    end
+
+    def properties config
+      output = {}
+      config['properties'].to_set.each do |prop|
```
sds: Is `to_set` necessary here? You can just iterate over the hash itself.

```ruby
config['properties'].each do |property, allowed_units|
   ...
end
```

sds:
Thanks for implementing my suggestions, @imcnally. Overall looks good, just a few nits I'd like to see cleaned up before merging. Thanks!

lib/scss_lint/linter/property_units.rb
imcnally: I'll think on that

lib/scss_lint/linter/property_units.rb
imcnally: I have such a love/hate thing with rubocop. Fixed in 12f9241.

lib/scss_lint/linter/property_units.rb
imcnally: Definitely wasn't. I made that harder than I needed to.

imcnally:
Thanks for taking the time @sds. All rubocop and test work was updated in 12f9241. Everything should be good to go!

config/default.yml
imcnally: You're right. I was confused by yml > ruby types.

spec/scss_lint/linter/property_units_spec.rb
imcnally: :bow:

sds:
Thanks @imcnally!

Merged in 52b52db. After having a second look, I noticed some issues with the implementation. I implemented some cleanup in f847d02 and fixed the issues in 71e45df and 7ad604b. The full set of changes can be viewed here: https://github.com/brigade/scss-lint/compare/52b52db...7ad604b

PR #408, opened 2015-04-03
Update the Rubies in .travis.yml

oliverklee:
- add Ruby 2.2
- always use the latest Ruby 2.1 instead of 2.1.0/2.1.1/2.1.2


sds:
Thanks!

Issue #335, opened 2015-01-13
Linter complains about valid BEM


martijnrusschen:
somehow the linter is complaining about this classname: `.table__row--align-top`

I configured the linter to enforce BEM class names, but according to this (http://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/) article it should. Any idea what's going wrong?

azizur:
Using this sample from http://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/:
```
.person {
  display: block;
}

.person__hand {
  display: block;
}

.person--female {
  display: block;
}

.person--female__hand {
  display: block;
}

.person__hand--left {
  display: block;
}

```
I am seeing:

```
[15:07:23] test.scss:9 [W] Selector `person--female` should be written in BEM (Block Element Modifier) format
[15:07:23] test.scss:13 [W] Selector `person--female__hand` should be written in BEM (Block Element Modifier) format
[15:07:23] test.scss:17 [W] Selector `person__hand--left` should be written in BEM (Block Element Modifier) format
```

Looks like we are doing BEM all wrong.

Anahkiasen:
No actually CSS Wizardry is advocating a BEM syntax which is not the standard (although that's the one I use too). To lint according to this article's syntax you need to set `convention` to `hyphenated_BEM`

azizur:
According to @csswizardry on http://cssguidelin.es/#bem-like-naming

The convention is:
> Elements are delimited with two (2) underscores (__), and Modifiers are delimited by two (2) hyphens (--).

According to BEM developers on (http://bem.info/tools/bem/bem-naming/):
> According to original BEM-naming convention it looks like the following:
>
> 'block[_block-mod-name[_block-mod-val]][__elem-name[_elem-mod-name[_elem-mod-val]]]'
>
> (Parameters whithin square brackets are optional)
>
>    Block — block-name.
>    Block's modifier in key-value format — block-name_mod-name_mod-val.
>    Block's boolean modifier — block-name_mod.
>    Block's element — block-name__elem-name.
>    Element's modifier in key-value format — block-name__elem-name_mod-name_mod-val.
>    Element's boolean modifier — block-name__elem_mod.

sds:
Duplicate of #319

PR #417, opened 2015-04-09
Fix modifier portion of BEM

conatus:
The current implementation does not understand the "modifier" portion of the BEM syntax.

I copied out the BEM regex as in the code to test it http://regexr.com/3apmp

As you can see adding a simple modifier in the standard BEM way doesn't work.

The implementation I have done here fixes it as the follow regex shows http://regexr.com/3apms

Thanks very much.


sds:
Hey @conatus,

I might be mistaken, but I believe the `hyphenated_BEM` syntax is what you are looking for? Please see the discussion in #335.

conatus:
Thanks for this - at least I made a PR to fix this "problem" amongst the other tickets!

Perhaps you could make this clearer in the documentation or alternatively rename BEM to "classic/strict/orthodox BEM" or something? Every tutorial I have seen, including some very recent ones uses the format that diverts slightly from the original formulation. The fact I wasn't the first to have this problem illustrates something is afoot.

sds:
That's a fair point, @conatus. You're definitely not the first to be tripped up by this and certainly won't be the last unless we do something.

I like your suggestion of renaming the convetion. I've chosen `strict_BEM` and made the change in 4d253c8337d8a99a69e87dbc1d90b23a1f15e089. Thanks!

conatus:
Thanks very much, this is cool.

Issue #426, opened 2015-04-17
Should VariableForProperty report lint for `transparent` and `inherit`?


chris-dura:
```yml
  VariableForProperty:
    enabled: true
    properties:
      - color
      - font
```
```scss
.foo {
   color: inherit;
   font: inherit;
}

.bar {
   color: transparent;
}
```
This reports lint, but I'm not sure it should?
Similar to my reasoning here: https://github.com/brigade/scss-lint/issues/376#issue-59166822

Writing this seems like a smell...
```scss
$inherit: inherit;
$transparent: transparent;

.foo {
   color: $inherit;
   font: $inherit;
}

.bar {
   color: $transparent;
}
```

sds:
Thanks for the suggestion, @chris-dura.

I've added a quick fix to this in 5a8cac9a2121c37b688712fc28f35e7aaa97dd36.

Can you think of other property values for which it might make sense to ignore? We may need to create a dictionary of values to ignore based on the property in question.

lencioni:
> Can you think of other property values for which it might make sense to ignore?

I added `currentColor` to the list: https://github.com/brigade/scss-lint/pull/433

PR #433, opened 2015-04-21
Ignore currentColor in VariableForProperty

lencioni:
Similar to 5a8cac9a2, this value is somewhat special and shouldn't be
considered as breaking the rules of this linter.

More information on the currentColor keyword can be found at:

  https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#currentColor_keyword

Addresses #426.


sds:
Thanks! Merged in f42486fbf1ba035ca4565498bad47521b9836140.

Issue #161, opened 2014-05-09
Add support for custom linters


marcogallotta:
Hey there

I am trying to add custom linters. The code we use to add our linters is included below. It was working in scss-lint version 0.12.0, but not on the latest version. We are enabling it by passing the args "-i <linter-name>", and the linters are in admin/lint/scss_lint_rules/ and follow the same structure as pre-packaged ones. It just never gets loaded into @options.

Any suggestions on how to get this working? We have also tried enabling it through a config file, but that doesn't seem to do anything.

```ruby
    require 'scss_lint'
  
    ASANA_LINT_DIR = "admin/lint/scss_lint_rules/"
  
    SCSSLint::LinterRegistry.class_eval do
      class << self
        def extract_linters_from(linter_names)
          puts "extracting linters"
          Dir["#{ASANA_LINT_DIR}*.rb"].each do |file|
            require file
          end
          linter_names.map do |linter_name|
            begin
              SCSSLint::Linter.const_get(linter_name)
            rescue NameError
              raise SCSSLint::NoSuchLinter.new("Linter #{linter_name} does not exist")
            end
          end
        end
      end
    end

    SCSSLint::CLI.new(ARGV).tap do |cli|
      puts "parsing args"
      cli.parse_arguments
      puts "running"
      cli.run
    end
```

and - any chance you'd add the functionality to pass custom linters in?  I'm sure lots of other groups have custom lint rules that don't make sense to add to this package, but would be of great use to them.

thanks!!

gunnx:
+1 for ForbiddenProperty

sds:
Experimental support for custom linters was implemented in #459. Please check out the `master` branch to take it for a spin. Feedback would be appreciated!

PR #439, opened 2015-04-26
Bliss::Module rule and default config

gilbox:
https://github.com/gilbox/css-bliss

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -200,5 +200,21 @@ linters:
   ZeroUnit:
     enabled: true
 
+  Bliss::Module:
+    enabled: false
+    severity: error
+    module_file_pattern: !ruby/regexp '/[\/\\]_?([A-Z][a-zA-Z0-9]+)\.scss/'
+    allow_id_selector_in_module: false
+    allow_attribute_selector_in_module: true
+    allow_element_selector_in_module: true
+
+    allow_module_margin: false
+    allow_module_width: false
+
+    allow_utility_classes_in_module: false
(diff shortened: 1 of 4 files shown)


sds:
Hey @gilbox, thanks for contributing to `scss-lint`!

This seems like a highly opinionated, relatively new framework for how to organize your CSS. That's fine&mdash;most of what `scss-lint` does is enabling developers to enforce opinionated rules&mdash;but it's suggesting to me that `scss-lint` really needs to provide a plugin system (sort of like what's described in #161) so developers can maintain these custom linters separately from the main `scss-lint` project.

Since @brigade doesn't subscribe to css-bliss, it's hard for us to justify taking it on and maintaining it as part of `scss-lint` proper. The same goes for `Compass`-related linters&mdash;we added this "section" of linters as an afterthought, but we don't use Compass ourselves internally and so this section doesn't really get any love or forward-moving development.

I think these are providing excellent reasons for why we should support a plugin-loading system for linters. This would allow developers such as yourself to maintain linters catered to certain frameworks, but still share them with a larger audience who can benefit from them.

I don't have a timeline for when this will get done, but I've been itching to implement plugin loading for a while now, and seeing the need here makes this even more appealing. The code you've written here will not go to waste, but you'll have to maintain a fork for the time being if you'd like to share this with others.

I'll be renaming the issue accordingly.

sds:
440

gilbox:
yep, I had read issue #161 which prompted me to open this PR. :+1: to supporting 3rd-party repos

PR #441, opened 2015-04-28
Closes #428; Adding a 'one_space_or_newline' style for SpaceAfterPropertyColon

### LINKED ISSUE #428: Linter config idea: Allow newline after property colon
Some organizations have a character limit for their SCSS lines. The only way I've seen this actually become a problem is with long property values, like URLs. So you might want to allow SCSS like:

```css
body {
  background-image:
      url('some-crazy-long-url-that-doesnt-fit-within-eighty-characters');
}
```

It would be nice to have a config style for SpaceAfterPropertyColon that would allow one space or a newline-and-spaces, but only if the `name: value;` is longer than _x_ characters.

I don't have a great idea what this style config should be called, or how to specify _x_, but maybe someone else has an idea... Maybe `style: one_space_or_newline_over_80_characters`, and the 80 is dynamically parsed out :worried: bleh.

srawlins:


sds:
Merged in f080325 with a fix to get the build passing in 9c27792.

PR #449, opened 2015-05-04
Fixes #442 by not Indentation linting an ImportNode if the previous node...

### LINKED ISSUE #442: Indentation lint is over-eager w/ multiple @import files
This one is kind of weak, but here me out :smile: 

a.scss

```scss
@import 'foo',
        'bar';

@import 'baz';

 @import 'xxx', 'yyy';

a {
  @include transition($property: opacity,
                      $timing: ease-in-out);
  background: inline-image('crazy-url.svg')
      center no-repeat;
}
```

```bash
$ bundle exec bin/scss-lint a.scss
a.scss:2 [W] Indentation: Line should be indented 0 spaces, but was indented 8 spaces
a.scss:6 [W] Indentation: Line should be indented 0 spaces, but was indented 1 spaces
a.scss:6 [W] Indentation: Line should be indented 0 spaces, but was indented 1 spaces
```

So what I would _not_ expect is: 1) the first lint, because a comma-separated list _feels_ like its all one node, starting at line 1, offset 0, and 2) the 3rd lint, for the same reason (here Indentation is double-reporting). My "comma-separated list is all one node" heuristic comes from the non-lints in the example, like the `@include` and `background` lines.

I have a fix for this, but it's unfortunately a little hacky: it depends on `previous_node()` which depends on a Node's `==`, which is ill-defined for `ImportNode`. So I have a monkey patch for that. The basic reason for the tricky/hacky fix is that the comma-syntax is just a shortcut; Sass creates a new `ImportNode` for each, and none of them contain the `@import` source.

srawlins:
... is an ImportNode that ends in a comma.

### DIFF
--- a/lib/scss_lint/linter/indentation.rb
+++ b/lib/scss_lint/linter/indentation.rb
@@ -89,6 +89,12 @@ module SCSSLint
       end
     end
 
+    def visit_import(node)
+      prev = previous_node(node)
+      return if prev.is_a?(Sass::Tree::ImportNode) && source_from_range(prev.source_range) =~ /,$/
+      check_indentation(node)
+    end
+
     # Define node types that increase indentation level
     alias_method :visit_directive, :check_and_visit_children
     alias_method :visit_each,      :check_and_visit_children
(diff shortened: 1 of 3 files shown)


sds:
Merged in 2b42faa04127e74d7be35b473d483594b88fc6c4.

PR #450, opened 2015-05-05
Add a linter that checks the number of elements in a BEM selector

justinlocsei:
This linter is disabled by default, allowing users to opt in to the linting if they are using BEM.  This defaults to allowing a single element, but can be configured to enable subelements by changing the
`max_elements` value.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -8,6 +8,10 @@ linters:
     space_before_bang: true
     space_after_bang: false
 
+  BemDepth:
+    enabled: false
+    max_elements: 1
+
   BorderZero:
     enabled: true
     convention: zero # or `none`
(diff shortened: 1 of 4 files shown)


justinlocsei:
Ah, right!  I'll get this fixed.

sds:
I'm fine with merging this for now, but if we start seeing more BEM-specific linters it may be worth thinking about how to extract them into a separate repository as discussed in #439. I have a feeling that will be difficult, however, given the number of linters with BEM-related options, but aren't strictly used for BEM.

Thanks for the pull request!

justinlocsei:
The Rubocop failures should now be fixed.  Thanks for considering this pull request, and for the great tool!

sds:
Merged in b4f6da97350aaebe782bd7450f5f3f7d9c1c0760.

PR #452, opened 2015-05-08
Added the ability to load config from the user's home directory

### LINKED ISSUE #436: Auto-load configuration file in home directory if no project-specific configuration found
Would it be possible to check for a .scss-lint.yml file in the user's home directory? So that it looks for config in the following order:

1. Working directory
2. Home directory
3. scss-lint config directory

For most linting tools (rubocop, jshint) I keep a config file in my home directory which is the default for all projects I work on, but it seems that scss-lint only looks in the working directory for config.

If you can give me a few directions I might be able to implement this myself.

iainbeeston:
This changes the way that config loading works, so that in addition to
checking in the current directory for a config file, scss-lint will also
check in the user's home directory, and load .scss-lint.yml from there
instead if it exists.

This changes the precedence of config loading to:

1. Command line argument
2. .scss-lint.yml in current directory
3. .scss-lint.yml in home directory
4. Config file packaged in the gem

In all cases only the config file loaded is used - config from multiple
files is not merged together.

Fixes #436

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Fix `Indentation` to not report `@import` directives spread across multiple
   lines
+* Add the ability to load config from the current user's home directory.
 
 ## 0.38.0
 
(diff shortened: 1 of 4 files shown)


sds:
Thanks for the pull request, @iainbeeston!

Merged in 006a272.

PR #455, opened 2015-05-13
Fixing YAML file so syck can parse it

madpilot:
The syck YAML parser vomits on the current default YAML file, because of the trailing square bracket in the PropertyUnits linter. Moving it up a level makes it happy.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -112,8 +112,7 @@ linters:
       'ms', 's',                               # Duration
       'Hz', 'kHz',                             # Frequency
       'dpi', 'dpcm', 'dppx',                   # Resolution
-      '%',                                     # Other
-    ]
+      '%']                                     # Other
     properties: {}
 
   PropertySortOrder:


sds:
Hey @madpilot,

Can you elaborate on what context we're parsing the default `scss-lint` YAML configuration with something other than the default YAML parser that ships with Ruby?

To me, this is a problem with the `syck` parser itself, not `scss-lint` (and I'd rather see an issue posted with that project). I'd like a better idea of why it is necessary for this file to be parse-able by `syck` without fixing the problem upstream.

madpilot:
I was following the instructions for including the default rake task in our Rails (3.2) app, which uses syck as the default YAML parser for legacy reasons. Since there isn’t a way to tell scss-lint to never read the default (it tries to merge), it was failing.

The simplest fix was to modify the YAML file, but a better(?) long term solution might be allow ignoring the default file completely?

> On 13 May 2015, at 2:03 pm, Shane da Silva <notifications@github.com> wrote:
> 
> Hey @madpilot,
> 
> Can you elaborate on what context we're parsing the default scss-lint YAML configuration with something other than the default YAML parser that ships with Ruby?
> 
> To me, this is a problem with the syck parser itself, not scss-lint (and I'd rather see an issue posted with that project). I'd like a better idea of why it is necessary for this file to be parse-able by syck without fixing the problem upstream.
> 
> —
> Reply to this email directly or view it on GitHub.
> 

----------------------------------------------
Myles Eftos
Mobile: +61-409-293-183

MadPilot Productions
URL: http://www.madpilot.com.au

Try our time tracking system: 88 Miles!
http://www.88miles.net

sds:
Alright, given the [`syck`](https://github.com/indeyets/syck) project hasn't seen a commit since 2012 this issue isn't likely going to get fixed.

Merged in 7f5badbb25b1d35791d1555eba11eddcda024fb5.

Issue #440, opened 2015-04-26
Allow linter plugins to be loaded from third-party repositories


sds:
See discussion in #439 for the context that prompted this. This will build on #161 to allow `scss-lint` to support a diverse collection of linters without having to directly support them in the mainline code. Interested parties can create their own plugins and developers can choose to extend `scss-lint` with those plugins if desired.

lencioni:
When considering how to architect the plugin system, it is probably worth looking at ESLint for inspiration: https://github.com/eslint/eslint

sds:
Experimental support for loading custom linters via gems was implemented in #459. Please check out the `master` branch to take it for a spin. Feedback would be appreciated!

PR #459, opened 2015-05-19
Allow configurable linter plugins

cih:
Hey Guys,

I noticed you had a couple of open issues that mentioned the need for supporting custom linters https://github.com/brigade/scss-lint/issues/161 and https://github.com/brigade/scss-lint/issues/440. I'm not sure exactly what you had in mind but I've had a go at implementing this to allow plugin directories and plugin gems. Used a couple of private methods as I didn't want to move stuff around too much without being familiar with the codebase, let me know what you think, hope this helps.

### DIFF
--- a/lib/scss_lint/config.rb
+++ b/lib/scss_lint/config.rb
@@ -178,6 +178,16 @@ module SCSSLint
       validate_linters
     end
 
+    def load_plugins
+      load_plugins_and_merge_config.tap { ensure_plugins_have_default_options }
+    end
+
+    def load_plugins_and_merge_config
+      Plugins.new(@options).load.each do |plugin|
+        @options = self.class.send(:smart_merge, plugin.config_options, @options)
+      end
+    end
+
     def enabled_linters
       LinterRegistry.extract_linters_from(@options['linters'].keys).select do |linter|
         linter_options(linter)['enabled']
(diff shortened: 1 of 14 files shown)


lib/scss_lint/config.rb
```diff
@@ -178,6 +178,16 @@ def initialize(options)
       validate_linters
     end
 
+    def load_plugins
+      load_plugins_and_merge_config.tap { ensure_plugins_have_default_options }
+    end
+
+    def load_plugins_and_merge_config
+      Plugins.new(@options).load.each do |plugin|
+        @options = self.class.send(:smart_merge, plugin.config_options, @options)
```
sds: The order that we're loading configurations here requires some closer attention. With this implementation, if I want to share a gem that sets configuration for linters included in the main `scss-lint` project (e.g. sets standard `Indentation` width or a maximum `NestingDepth`) that can be used across all projects in my organization, I can't.

This is because the default configuration that is automatically loaded with every configuration will override the plugin configuration here.

This suggests that this plugin loading code needs to be shimmed somewhere in the actual configuration loading process _before_ the default configuration is merged. This preserves the load order of `default` → `plugins` → `local .scss-lint.yml`. I've implemented this in a382abf.

sds:
Thanks for the pull request @cih!

I've merged in in 55505e5 with some adjustments in 1d9d7ee...9b6307c.

I've very excited for this feature, but it would be great if we could get some people trying it out. This represents a big step in the `scss-lint` project as it must now start to support an "official" (as opposed to private) API for writing linters.

I've no doubt we'll discover some issues with this new feature. I'm hoping the community will help us discover and fix them as we go.

PR #466, opened 2015-05-26
Fix sort order link in linter docs

istro:
Link in linter docs appears to be from the times when linter readme was part of main project readme. Not a fan of absolute path, but it's better than a 404.

### DIFF
--- a/lib/scss_lint/linter/README.md
+++ b/lib/scss_lint/linter/README.md
@@ -810,7 +810,7 @@ execute `:sort` in `vim`), and it can
 You can also specify an explicit ordering via the `order` option, which allows
 you to specify an explicit array of properties representing the preferred
 order, or the name of a
-[preset order](data/property-sort-orders).
+[preset order](../../../../master/data/property-sort-orders).
 If a property is not in your explicit list, it will be placed at the bottom of
 the list, disregarding its order relative to other unspecified properties.
 


lib/scss_lint/linter/README.md
```diff
@@ -810,7 +810,7 @@ execute `:sort` in `vim`), and it can
 You can also specify an explicit ordering via the `order` option, which allows
 you to specify an explicit array of properties representing the preferred
 order, or the name of a
-[preset order](data/property-sort-orders).
+[preset order](https://github.com/brigade/scss-lint/tree/master/data/property-sort-orders).
```
sds: Would a [relative link](https://github.com/blog/1395-relative-links-in-markup-files) work here? 

e.g. `[preset order](../data/property-sort-orders)`

lib/scss_lint/linter/README.md
```diff
@@ -810,7 +810,7 @@ execute `:sort` in `vim`), and it can
 You can also specify an explicit ordering via the `order` option, which allows
 you to specify an explicit array of properties representing the preferred
 order, or the name of a
-[preset order](data/property-sort-orders).
+[preset order](../../../../master/data/property-sort-orders).
```
istro: it sorta does, but it's sorta ugly this way, too. I almost prefer the hardcoded link, it's at least readable at a glance.

lencioni:
Thanks for the PR! This was fixed by #469.

PR #467, opened 2015-05-26
add a few properties to smacss property order preset

istro:
Smacss book seems to have just a general idea about property order, and I feel this one you have as a preset could have a couple more properties.

Added `box-sizing`, `animation`, `list-style-type` and `list-style-image`, `word-wrap` and `word-break`.

Grouped `letter-spacing` and `word-spacing` together. 

Removed `text-wrap` as it appears to be [deprecated](http://www.w3schools.com/cssref/css3_pr_text-wrap.asp).

No urgency, not a huge deal really as setting `ignore_unspecified: true` helps with most of these - but hey, might as well suggest improving the predefined config.

### DIFF
--- a/data/property-sort-orders/smacss.txt
+++ b/data/property-sort-orders/smacss.txt
@@ -30,6 +30,8 @@ padding-right
 padding-bottom
 padding-left
 
+box-sizing
+
 float
 clear
 
@@ -43,6 +45,7 @@ column-width
 
 transform
 transition
+animation
 
 # Border
 
@@ -102,9 +105,10 @@ font-style
 font-variant
 font-weight
 
-letter-spacing
 line-height
 list-style
+list-style-type
+list-style-image
 
 text-align
 text-decoration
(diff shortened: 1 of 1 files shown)


sds:
Closing due to lack of response from @istro, specifically on the issue of why `letter-spacing` was separated from other text-related properties.

This raises an important point: I don't think it makes sense for `scss-lint` to be the discussion ground for third-party sorting presets, especially when those presets are not part of a published standard. Hopefully the work done in #459 makes it easier for individuals to "own" the SMACSS sort ordering. They could simply create a repository with a SMACSS sort order specified in the configuration for `PropertySortOrder.order` option.

istro:
That was quick "lack of response" :-)

In any case, I do agree with the point addressed in PR 459. This is better yet :+1:. I might be one of the people trying out the new feature :japanese_ogre:

As to @snookca's comment - makes sense, too. Then `word-spacing` being text-related also belongs in that same group with `line-height` and `letter-spacing`, while list style ones should be separate I guess.

PR #484, opened 2015-06-08
Add box-sizing to smacss

danieldiekmeier:


danieldiekmeier:
I hope this is the right place for it. I also thought about putting it near the top, between `display` and `position`?

sds:
Since we at @brigade don't subscribe to the SMACSS philosophy per se, I can't speak to whether this would be an appropriate place to order `box-sizing`.

Now that an implementation for #440 has landed on `master`, I'm hoping someone from the community can step forward and "own" any SMACSS related configuration, so we can remove this default ordering from `scss-lint` (and developers who want it can load the plugin separately).

Until #440 gets released, can anyone else who uses SMACSS vouch for this ordering?

danieldiekmeier:
Would it also be possible to add support for something like `order: 'my-ordering.txt'`, so I don't need to have the whole CSS definition inside my yml?

danieldiekmeier:
I take that back.

Issue #435, opened 2015-04-22
Re-add support to lint content not available as a file


teoljungberg:
Before 0.35.0 we could send in strings, an array of strings to be more precise,
into `SCSSLint::Runner#run` and get back the violations for that file. But,
since 0.35.0 shipped, all arguments given to `SCSSLint::Runner#run` must be
actual files. Sadly we don't have access to the actual files in [HoundCI](https://houndci.com) which
locks us into to not being able to upgrade to 0.35.0 and onwards.

Could we re-add support to being able to lint content of that does not come from files?

teoljungberg:
I'll gladly help out doing this btw

sds:
Hey @teoljungberg,

I didn't realize HoundCI was using `scss-lint` in this manner. Can you elaborate on why passing the file directly is not possible?

The issue here is that the entire linting framework assumes a lint is associated with a file, so your lint warnings/errors will have to have some dummy filename in them if you're going to get this to work. Why not just write a temporary file?

I suppose I'm not against adding support for this (I didn't realize this was possible before, so I'm not sure what exactly changed), but I'm wary of introducing a hack for a use case easily solved by passing through temporary files.

teoljungberg:
We could re-write the actual linting to use temporary files, but I think it would be nicer to support both actual files and content of files. Since it seems to be baked in https://github.com/brigade/scss-lint/blob/b0de02f59ec5c376e2e5d73e673ffea96e9dd826/lib/scss_lint/engine.rb#L20-L24

sds:
That functionality was added so we can easily write specs for linters, but you're right that it shouldn't be too bad to support files and raw strings.

I'd like to have the `Runner` class act as the main point of contact between HoundCI and `scss-lint`'s engine. I believe the best way to support what you're asking is to add support to `Runner#run` to accept an options hash that takes either `files` or `sources` as a key, and then have the `Runner` do the rest. Make sense?

teoljungberg:
I see, that makes sense.

Having `Runner` being the consumed API sounds like a good idea, that's
what we're using in Hound anyhow. I like the interface of `files` and
`sources` aswell.

franzheidl:
+1 for support of linting strings.
Would be great to use scss-lint for things like text editor/ide plugins to lint the current contents of a file, no matter whether the current state of the file is saved or not.

teoljungberg:
@sds Has there been any movement on this subject?

sds:
[...]

PR #491, opened 2015-06-12
Lint string aswell as files

teoljungberg:
References #435

### DIFF
--- a/lib/scss_lint/runner.rb
+++ b/lib/scss_lint/runner.rb
@@ -22,7 +22,8 @@ module SCSSLint
 
     # @param file [String]
     def find_lints(file)
-      engine = Engine.new(file: file)
+      options = engine_options_for(file)
+      engine = Engine.new(options)
 
       @linters.each do |linter|
         begin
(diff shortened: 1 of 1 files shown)


teoljungberg:
I'm unsure how a test-case for this should look, any ideas @sds ?

sds:
Hey @teoljungberg, thanks for the pull request.

The reason I don't want to revert back to the previous behavior of testing for file existence to determine whether to treat a string as code or a file is because it's not terribly clean.

For example, while perhaps a silly example, the string `p{}` is a valid file name as well as valid SCSS code. Do we prioritize loading the file over parsing it as code in this case? (again, we can argue about whether this is practical, but it is an issue).

I was hoping we could adjust the runner to instead accept an options hash with `files` or `source` keys, and then lint based on whichever keys are present. This changes the signature, but I think is a cleaner approach if we want to support this in a proper way.

teoljungberg:
> I was hoping we could adjust the runner to instead accept an options hash with files or source keys, and then lint based on whichever keys are present. This changes the signature, but I think is a cleaner approach if we want to support this in a proper way.

@sds - That's sounds like a good approach. I wanted to make the smallest imprint at first to get this kind of feedback. I'll update the PR with the comments you made

tobiashm:
Maybe if the input was just a stream, which might come from a file, but could also come from `stdin` or in the form of `StringIO`.

I would really like the CLI to be able to accept a `stdin` stream, as that would make integration with other tools, like editors, much easier.

sds:
Closing this pull request as it seems to have been abandoned. Feel free to open a new one with the [comments above](https://github.com/brigade/scss-lint/pull/491#issuecomment-112192160) addressed. Thanks!

PR #495, opened 2015-06-18
Control comment fixes

ivantsepp:
Several fixes for control comment functionality

1: When control comment appears in the middle of a comma sequence
This was introduced by 1d892e2a. Looks like the tests are false positive because the rule is the only child of the root node. The following should raise a lint error but doesn't:
```
.a, // scss-lint:disable LeadingZero
.b {
  width: 0.5px;  
}
.another {
  color: #FFF;
}
```

2: `FinalNewline`, `SpaceBetweenParens`, and `TrailingWhitespace` linters add the lint errors in the `visit_root` method but it doesn't `yield` for the other nodes to be visited. That means any top level comment to disable all will not work. The following still raises lint errors:
```
// scss-lint:disable all
.another {
  color: rgba( 0, 0, 0, .1 );
}
```

3: Finally, for `FinalNewline` and `TrailingWhitespace`, you can get lint errors if your last line is the trailing `}`. The following will raise lint errors when it shouldn't:
```
// scss-lint:disable all
.another:after {
  content: 'There is no new final line and trailing spaces after the }'
}    
```

### DIFF
--- a/lib/scss_lint/control_comment_processor.rb
+++ b/lib/scss_lint/control_comment_processor.rb
@@ -82,10 +82,21 @@ module SCSSLint
       return unless comment_node = @disable_stack.pop
 
       start_line = comment_node.line
+      if comment_node.class.node_name == :rule
+        end_line = start_line
+      elsif node.class.node_name == :root
+        end_line = @linter.engine.lines.length
+      else
+        end_line = end_line(node)
+      end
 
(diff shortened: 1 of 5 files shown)


sds:
Thanks for the fix, @ivantsepp!

Merged in 1d591fac254efc43b4746746e01038fe5bc1f521.

PR #508, opened 2015-07-01
Reorder config to be alphabetical & match documentation

edhenderson:
- easier to follow along if default config and documentation are in the same order


sds:
Thanks! Merged in be6c30fa2f7df4166473a5cf84c3f0decce5ead0 after fixing some trailing whitespace and commit message nits.

PR #509, opened 2015-07-01
New linter: SpaceAroundOperator

### LINKED ISSUE #400: Proposed linter: space around operators
It would be nice to have a linter that ensures there are spaces around operators. This could work similarly to ESLint's space-infix-ops rule: http://eslint.org/docs/rules/space-infix-ops.html

Example bad:

```scss
padding: $some-padding*2;
```

Example good:

```scss
padding: $some-padding * 2;
```

srawlins:
Fixes #400 

Essentially explained in the README change, and in [my comment on #400](https://github.com/brigade/scss-lint/issues/400#issuecomment-116768105).

I think that `#all_matches` is a weak spot in this PR: it's not well unit-tested. However, it's not general purpose enough to go into `lib/scss-lint/utils.rb` (which also doesn't have it's own unit tests), and the specs revealed a few bugs that I fixed along the way. If this method should be implemented in some other way, I'm open.

### DIFF
--- a/lib/scss_lint/linter/space_around_operator.rb
+++ b/lib/scss_lint/linter/space_around_operator.rb
@@ -5,62 +5,44 @@ module SCSSLint
 
     def visit_script_operation(node)
       source = source_from_range(node.source_range)
+      left_range = node.operand1.source_range
+      right_range = node.operand2.source_range
 
-      all_matches(source) do |match|
-        if config['style'] == 'one_space'
-          if match[:lspace] != ' ' || match[:rspace] != ' '
-            add_lint(node, space_message(match))
-          end
-        elsif match[:lspace] != '' || match[:rspace] != ''
-          add_lint(node, no_space_message(match))
+      # We need to #chop at the end because an operation's operand1 _always_
+      # includes one character past the actual operand (which is either a
+      # whitespace character, or the first character of the operation).
+      left_source = source_from_range(left_range).chop
+      right_source = source_from_range(right_range)
+      operator_source = source_between(left_range, right_range)
+
+      match = operator_source.match(%r{
+        (?<left_space>\s*)
+        (?<operator>\S+)
+        (?<right_space>\s*)
+      }x)
+
+      if config['style'] == 'one_space'
+        if match[:left_space] != ' ' || match[:right_space] != ' '
+          add_lint(node, SPACE_MSG % [source, left_source, match[:operator], right_source])
         end
+      elsif match[:left_space] != '' || match[:right_space] != ''
+        add_lint(node, NO_SPACE_MSG % [source, left_source, match[:operator], right_source])
       end
+
+      yield
     end
 
   private
+    SPACE_MSG = "`%s` should be written with a single space on each side of the operator: `%s %s %s`"
+    NO_SPACE_MSG = "`%s` should be written without spaces around the operator: `%s%s%s`"
 
-    OPERATOR_REGEX = /
-      \b
-      (?<operation>
-        (?<loperand>[^\s]+?)
-        (?<lspace>\s*)
-        (?<operator>
-          (?:==|!=|>|>=|<|<=) |  # Comparison operators
-          (?:[+\-*\/%])          # Mathematical operators
-        )
-        (?<rspace>\s*)
-        (?<roperand>[^\s]+)
-      )
-      \b
-    /ix
+    def source_between(range1, range2)
+      # We don't want to add 1 to range1.end_pos.offset because of the same
+      # reason as the #chop comment above.
+      between_start = Sass::Source::Position.new(range1.end_pos.line, range1.end_pos.offset)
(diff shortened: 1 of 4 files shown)


sds:
Hey @srawlins, thanks for tackling this feature!

It looks like this returns false positives for any variables with dashes in the name, e.g. `$my-variable` (these are valid variable names in Sass).

Can I get you to fix that issue? Other than that this looks great.

srawlins:
Hi sds@. This shouldn't be the case; I added a test case though to track it.

sds:
Sorry for the lack of example, @srawlins.

When running your linter I get false positives for dashed variables in expressions, e.g.

```scss
p {
  font-size: em(16px / $base-font-size);
}
```

srawlins:
Great catch, @sds! That example kind of revealed that my implementation was garbage.

I totally re-implemented SpaceAroundOperator, and its shorter, cleaner, and restricts itself to the operators that Sass has identified. I added one more test as well with newlines.

sds:
Looking much better. Still seems to be some unexpected results when it comes to parentheses. A silly example:

```scss
div {
  top: ($padding) * -1;
}
```

...outputs:

```
`$padding) * -1;` should be written with a single space on each side of the operator: `$padding ) -1;`
```

srawlins:
Phhhhhhhhh another good catch, @sds ... This one is tough. I dunno why Sass doesn't parse the parens... differently. It's tougher because there could be more parens and whitespace, like `top: (($padding   ) ) * -1`. Hrm... I'll play with this.

srawlins:
OK the fix wasn't as bad as I thought. This is ready again.

The fix assumes that this parentheses is a weird special case of only the right parenthesis character, `)`, shifting them off the operand, onto the left operator, before evaluating whether there is lint.

lib/scss_lint/linter/space_around_operator.rb
```diff
@@ -0,0 +1,57 @@
+module SCSSLint
+  # Checks for space around operators on values.
+  class Linter::SpaceAroundOperator < Linter
+    include LinterRegistry
+
+    def visit_script_operation(node)
+      source = source_from_range(node.source_range)
```
sds: I propose changing to:

```ruby
source = source_from_range(node.source_range).gsub(/\)\z/, '')
```

to remove trailing parentheses from the lint message.

lib/scss_lint/linter/space_around_operator.rb
```diff
@@ -0,0 +1,57 @@
...
+      left_range = node.operand1.source_range
+      right_range = node.operand2.source_range
+
+      # We need to #chop at the end because an operation's operand1 _always_
+      # includes one character past the actual operand (which is either a
+      # whitespace character, or the first character of the operation).
+      left_source = source_from_range(left_range).chop
+      right_source = source_from_range(right_range)
```
sds: Similarly, changing this to:

```scss
right_source = source_from_range(right_range).gsub(/\)\z/, '')
```

...will remove trailing parentheses if they exist, making the lint message easier to read.

sds:
Thanks for continuing to dig into this, @srawlins.

Some additional odd cases:

```scss
@if $var == 1 or $var == 2 {
...
}
```

```
`$var == 1 or $var == 2 {` should be written with a single space on each side of the operator: `$var == 1  or $var == 2 {`
```

...and:

```scss
p {
  margin: $var + $var2 + $var3;
}
```

```
`$var + $var2 + $var3;` should be written with a single space on each side of the operator: `$var + $var2  + $var3;`
```

Both these seem to illustrate the same problem, where the space separating the nested left expression from the right expression is not recognized.

lib/scss_lint/linter/space_around_operator.rb
srawlins: I added `chop` instead, which is working in all the examples I have (removing semicolons, right parens, left curly braces...).

srawlins:
Thanks for testing, @sds , it looks like this always happens when the left operator is an operation itself (as opposed to a simple value). I've fixed, and added more examples.

srawlins:
Oops, looks like one spec now fails on Sass 3.4.16 ([Travis](https://travis-ci.org/brigade/scss-lint/builds/71117379)), but this spec does not fail for me on Sass 3.4.15. From Sass's [CHANGELOG](https://github.com/sass/sass/blob/stable/doc-src/SASS_CHANGELOG.md#3416-10-july-2015):

> Parse a minus operator with no surrounding whitespace (e.g. 20px-10px) the same regardless of whether it's being used in a static property.

So newly in 3.4.16, `margin: 2em-1em;` has an operation in it. Unfortunately this _should_ be a lint in 3.4.16, and _should not_ be a lint <= 3.4.15, so I've just commented out the example for now.

sds:
Sorry for the delay in response, @srawlins.

Merged in c44e3e5 with an additional fix for multiline expressions in c8c6a99. Thanks for your hard work on this!

PR #519, opened 2015-07-09
New linter: DisableLinterReason

### LINKED ISSUE #514: Linter Idea: DisableLinterReason
This one might be far-fetched, but maybe not. When disabling linters inline, as documented in the [README](https://github.com/brigade/scss-lint/#disabling-linters-via-source), I want to see a comment above the `scss-lint:disable` instruction that explains why individual linter(s) are being disabled via source. Therefore:

**Bad**

```scss
// scss-lint:disable BorderZero
p {
  border: none;
}
```

**Good**

```scss
// This file uses `border: none` for reasons and things. See http://some-discussion.
// scss-lint:disable BorderZero
p {
  border: none;
}
```

One implementation difficulty would be that I think Sass parses these lines into **one** comment node, not two. Hrm...

srawlins:
Fixes #514. DisableLinterReason is explained over at #514 as well.

As for the implementation, a few notes:

* I chose to disable it by default, since it's not exactly standard, or obvious.
* I copied the `COMMAND_REGEX` from `control_comment_processor.rb`. There's probably a better way, but I dunno what y'all would prefer. Make `COMMAND_REGEX` a public constant in `control_comment_processor.rb`? Extract it out to `constants.rb`? `utils.rb`?

### DIFF
--- a/lib/scss_lint/linter/disable_linter_reason.rb
+++ b/lib/scss_lint/linter/disable_linter_reason.rb
@@ -4,9 +4,9 @@ module SCSSLint
     include LinterRegistry
 
     def visit_comment(node)
-      return unless node.value.first.match(COMMAND_REGEX)
-
-      # No lint if the first line of the comment is not the command.
+      # No lint if the first line of the comment is not a command (because then
+      # either this comment has no commands, or the first line serves as a the
+      # reason for a command on a later line).
       return unless comment_lines(node).first.match(COMMAND_REGEX)
 
       # Maybe the previous node is the "reason" comment.
--- a/lib/scss_lint/linter/disable_linter_reason.rb
+++ b/lib/scss_lint/linter/disable_linter_reason.rb
@@ -27,7 +27,7 @@ module SCSSLint
     COMMAND_REGEX = %r{
       (/|\*)\s* # Comment start marker
       scss-lint:
-      (?<action>disable|enable)\s+
+      (?<action>disable)\s+
       (?<linters>.*?)
       \s*(?:\*/|\n) # Comment end marker or end of line
     }x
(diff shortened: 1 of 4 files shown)


lib/scss_lint/linter/disable_linter_reason.rb
```diff
@@ -0,0 +1,39 @@
+module SCSSLint
+  # Checks for "reason" comments above linter-disabling comments.
+  class Linter::DisableLinterReason < Linter
+    include LinterRegistry
+
+    def visit_comment(node)
+      return unless node.value.first.match(COMMAND_REGEX)
```
sds: This will also require comments for the `enable` counterpart, so for code like the following:

```scss
// scss-lint:disable ...
p {
}
// scss-lint:enable ...
```

...I would get two lints. Perhaps we should only report disabling without a reason?

lib/scss_lint/linter/disable_linter_reason.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      # No lint if the first line of the comment is not the command.
+      return unless comment_lines(node).first.match(COMMAND_REGEX)
```
sds: Is the check above necessary with this present, or are they testing something different?

srawlins: Good catch, I removed the first check, and made the comment more clear.

lib/scss_lint/linter/disable_linter_reason.rb
srawlins: I agree. Ok this only lints on `scss-lint:disable` now.

sds:
Thanks for the pull request!

Merged in 3f459fe.

PR #525, opened 2015-07-14
Fixing non-multiple match failure message, and tons of matcher tests

### LINKED ISSUE #524: report_lint matcher: Failed match of specific count of lints prints a useless message
In the linter specs, we can specify that we expect an exact number of lints on a line (this is used a number of times; `git grep 'line: [0-9]*, count: [0-9]*' spec/scss_lint/linter`), but when it fails the message is not useful.

For example, if we change border_zero_spec.rb, line 41, to: `it { should report_lint line: 2, count: 2 }`, then we get the following failure from RSpec:

```
1) SCSSLint::Linter::BorderZero when a property has a border of none should report a lint on line 2
   Failure/Error: it { should report_lint line: 2, count: 2 }
     expected that a lint would be reported on line 2, but was on line 2
   # ./spec/scss_lint/linter/border_zero_spec.rb:41:in `block (4 levels) in <top (required)>'
```

That is, "expected that a lint would be reported on line 2, but was on line 2".

srawlins:
Fixes #524

When fixing #524, I thought that the message logic is complicated enough to warrant testing. So I wrote a crazy long test for `report_lint`.

### DIFF
--- a/spec/scss_lint/linter/bang_format_spec.rb
+++ b/spec/scss_lint/linter/bang_format_spec.rb
@@ -105,7 +105,7 @@ describe SCSSLint::Linter::BangFormat do
     SCSS
 
     it 'does not loop forever' do
-      subject.should_not report_lint
+      should_not report_lint
     end
   end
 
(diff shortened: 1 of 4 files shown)


sds:
Merged in 2062370304ce046138660a028e6cbb94b01e85b1. Thanks!

Issue #528, opened 2015-07-17
PropertySortOrder fails on any non-default ordering of flex properties


jakezatecky:
I have come across an annoying issue where the `PropertySortOrder` linter always throws an error regardless of any ordering of a specific combination of flex properties when using a non-default ordering mechanism.

The `.scss-lint.yml` file can have any order besides the default lexicographical order (that is, either concentric/recess/smacss):

``` yml
linters:
  PropertySortOrder:
    enabled: true
    ignore_unspecified: false
    separate_groups: false
    order: recess
```

These are the properties that, when existing in the same rule, will trigger the error. The subtraction of any one of these properties will remove the in-escapable error-loop created by `PropertySortOrder`:

``` css
.flex {
  display: flex;
  flex-direction: row;
  align-items: center;
  justify-content: flex-end;
}
```

The ordering above produces the following error:

```
[W] PropertySortOrder: Properties should be ordered display, align-items, flex-direction, justify-content
```

If I re-order the properties in the suggested manner, I get the following error:

```
[W] PropertySortOrder: Properties should be ordered display, flex-direction, align-items, justify-content
```

As you can see, it reports a different ordering when you switch to the recommended ordering! I also tried to sort the properties lexicographically, but that produced an error as well.

There are ways around this, of course (such as disabling the linter), but that is not optimal.

srawlins:
I think this was fixed in 0.40.0, specifically in 04f22b2faa5e65ce1a10916d298d8a63bf3b4121.

sds:
I'm unable to reproduce, and I agree this looks suspiciously like #478. 

Closing as duplicate.

PR #529, opened 2015-07-17
Add the flexbox properties to the recess sort

srawlins:
Mitigates #528 

Unfortunately, Recess seems like a dead project, last updated in 2013. Someone has an open pull request though, to add the flexbox properties (twitter/recess#148). I snagged the file, selected the order array, transformed it, and popped it into scss-lint's recess.txt with the following shell:

```bash
$ wget https://github.com/ImBobby/recess/raw/master/lib/lint/strict-property-order.js
$ echo "# RECESS Property Order" > data/property-sort-orders/recess.txt
$ echo "# https://github.com/twitter/recess" >> data/property-sort-orders/recess.txt
$ echo "" >> data/property-sort-orders/recess.txt
$ cat strict-property-order.js \
    | awk '/order = \[/,/\]/' \
    | awk "/'/" \
    | sed "s/.*'\(.*\)'/\1/" >> data/property-sort-orders/recess.txt
```

Note: There is certainly a bug in PropertySortOrder. Even with missing properties, @jakezatecky should not experience the two different recommended sort orders. But this PR at least mitigates the problem for anyone using `order: recess` and flex properties.


sds:
Thanks @srawlins.

Merged in 1fc6020fadcfa30c976dbf89e7b0564ba6c87a10.

Issue #343, opened 2015-01-16
Auto-correcting


Dorian:
Most things can be autocorrected (eg: colors, paddings, margins, order of rules, etc...)

Similar to [rubocop](https://github.com/bbatsov/rubocop)'s `--auto-correct`

> -a/--auto-correct	Auto-correct certain offenses. Note: Experimental - use with caution.

Anahkiasen:
Would really love this, in particular for properties order which is really painful and long to correct on big codebases

Dorian:
Yep, I think with CSS 95% of the "warnings" I got are auto-correctable (eg: the "line too longs", "too many nested rules", ... are impossible to auto-correct automaticaly

sds:
This is far from trivial, and so I don't think it is a worthwhile pursuit in `scss-lint`.

The `rubocop` project enjoyed this capability thanks to the underlying `parser` gem that they use. There is no matching capability available with the `sass` parser, and without that this sort of feature becomes an entire project unto itself.

Thus, as mentioned in #282, unless some very enthusiastic volunteer implements a custom Sass parser with tree mutation capability, you will not see this in `scss-lint`.

Hope that makes sense.

Dorian:
Hmm, yes right I think this should be another project, I might a simple regexp based + basic rule sorting one.

Dorian:
Actually I started a repo with some basic regexp based scripts: https://github.com/Dorian/scss-lint-auto-correct

PR #534, opened 2015-07-25
Support to auto correct a lint

brauliobo:
FIXME: broke monkey patch at lib/scss_lint/sass/tree.rb

### DIFF
--- a/lib/scss_lint/cli.rb
+++ b/lib/scss_lint/cli.rb
@@ -54,9 +54,12 @@ module SCSSLint
     end
 
     def scan_for_lints(options, config)
+      files = FileFinder.new(config).find(options[:files])
+
       runner = Runner.new(config)
-      runner.run(FileFinder.new(config).find(options[:files]))
+      engines = runner.run(files)
       report_lints(options, runner.lints, runner.files)
+      auto_correct engines if options[:auto_correct]
 
       if runner.lints.any?(&:error?)
         halt :error
@@ -166,6 +169,15 @@ module SCSSLint
         results = reporter.new(sorted_lints, files).report_lints
         io = (output == :stdout ? $stdout : File.new(output, 'w+'))
         io.print results if results
+      end
+    end
+
+    def auto_correct(engines)
+      runner.lints.each do |lint|
+        lint.auto_correct.call if lint.auto_correct
+      end
+      engines.each do |engine|
+        puts engine.tree.to_scss
       end
     end
 
--- a/README.md
+++ b/README.md
@@ -80,6 +80,7 @@ Command Line Flag         | Description
 `-r`/`--require`          | Require file/library (mind `$LOAD_PATH`, uses `Kernel.require`)
 `-i`/`--include-linter`   | Specify which linters you specifically want to run
 `-x`/`--exclude-linter`   | Specify which linters you _don't_ want to run
+`-a`/`--auto-correct`     | Auto correct files and print them to STDOUT (currently only a few Linters support this)
 `-h`/`--help`             | Show command line flag documentation
 `--show-formatters`       | Show all available formatters
 `--show-linters`          | Show all available linters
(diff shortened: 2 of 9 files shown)


brauliobo:
the intention was not the current API, so the change was quite small to achieve the apply behaviour

brauliobo:
change the monkey path to be less intrusive and all tests pass now

sds:
Hey @brauliobo,

There's a lot going on in this pull request without much documentation. From what I see, you are trying to do four separate things:

* Replace `scss-lint`'s monkey patching of `Node#children` with a separate `linter_children` call
* Add support for automatically fixing lints? (the `--apply` flag name isn't very descriptive here. Other tools usually use the name `--auto-correct`)
* Split `MergeableSelector` into separate linters (Why? An explanation of the purpose for this would be useful, as it introduces a backwards incompatibility with developers who configure `MergeableSelector`)
* Add a new `GroupableSelector` linter (again, with no documentation--please add documentation to the [linter README](https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md))

Please split these out into separate pull requests with clear descriptions of the problems they are trying to solve. This allows us to merge them individually without holding up the whole set due to problems/concerns with one of them.

Furthermore, you've probably noticed that the lint checks are not passing. Please ensure all RuboCop-related warnings are addressed before submitting a pull request. Thanks!

brauliobo:
@sds, nice, gonna follow your advices, hope I can find some time soon.

brauliobo:
@sds applied your suggestions

The first two suggestions is the this PR, as the Sass::Tree fix is core dependency for auto correcting
Just created the #535 and #536 for the last two.

Still there is Rubocop issues, sorry for this, I'm quite not used to it, will get used to it and fix the issues.

brauliobo:
fixed rubocop warnings

lib/scss_lint/cli.rb
```diff
@@ -54,9 +54,12 @@ def act_on_options(options)
     end
 
     def scan_for_lints(options, config)
+      files = FileFinder.new(config).find(options[:files])
+
       runner = Runner.new(config)
-      runner.run(FileFinder.new(config).find(options[:files]))
+      engines = runner.run(files)
```
sds: This is not an ideal approach. Calling `run` should not return a list of engines. The runner should auto-correct if auto-correction has been called for. There are a lot of other issues with the `Runner` class that are unrelated to your change, but it's purpose is to be the central coordinator for the tool. The `CLI` class is simply intended to parse CLI options and hand those off to the `Runner`.

lib/scss_lint/cli.rb
```diff
@@ -169,6 +172,15 @@ def report_lints(options, lints, files)
       end
     end
 
+    def auto_correct(engines)
+      runner.lints.each do |lint|
+        lint.auto_correct.call if lint.auto_correct
+      end
+      engines.each do |engine|
+        puts engine.tree.to_scss
```
sds: I don't think this works the way you think it does. Calling `to_scss` is likely to change more of the original source code than you originally intended due to the quirks of the Sass engine. 

Furthermore, using `puts` to output the results prevents us from controlling the file we output to via p[`--out` flags](https://github.com/brigade/scss-lint/pull/273). A better solution would simply update the file at the end of the run.

However, the big show-stopper here is using `to_scss`, which will butcher the original source tree.

sds:
Hey @brauliobo, thanks for the pull request. Really appreciate the effort to add support for something that I know a lot of users are looking for.

Sadly, I'm concerned with the general approach here. Relying on `to_scss` is not a viable option, as it will re-serialize the entire source code tree. As I've already explained in https://github.com/brigade/scss-lint/issues/343#issuecomment-70279869, other projects (such as `rubocop`) benefit from a better parse tree which can actually be changed at runtime and re-serialized without touching the rest of the source code. We don't have that flexibility with the `sass` gem here, so any solution will need to implement that from scratch (a huge undertaking, IMO).

sds:
Closing due to lack of response. Feel free to reopen if you can come up with a new approach!

PR #535, opened 2015-07-28
Split MergeableSelector into MergeEqualSelector and NestableSelector

brauliobo:
Depends on #534 

MergeableSelector can't handle two different cases like this, as `equal?` condition has priority over the `nested?` then eventually nesting cases won't be handled.


config/default.yml
```diff
@@ -92,9 +92,11 @@ linters:
     enabled: true
     style: exclude_zero # or 'include_zero'
 
-  MergeableSelector:
+  MergeEqualSelector:
+    enabled: true
+
+  NestableSelector:
     enabled: true
-    force_nesting: true
```
sds: Can you explain why this option is no longer necessary?

lib/scss_lint/linter/merge_equal_selector.rb
```diff
@@ -0,0 +1,60 @@
...
+  class Linter::MergeEqualSelector < Linter
+    include LinterRegistry
+
+    def check_node(node)
+      visits = Set.new
+      node.children.each do |child_node|
+
+        equal_node = find_equal_node child_node, node.children
+        next unless equal_node
+        next if visits.include? [child_node, equal_node] or
```
sds: The verbal `or`/`and` boolean operators aren't used anywhere else in this repository, and we'd like to keep it that way due to the way their precedence is handled. Please use `||`/`&&` instead.

lib/scss_lint/linter/merge_equal_selector.rb
```diff
@@ -0,0 +1,60 @@
...
+                visits.include? [equal_node, child_node]
+        visits << [child_node, equal_node]
+
+        add_lint child_node.line,
+        "Merge rule `#{node_rule(child_node)}` on line #{child_node.line} with rule " \
+        "on line #{equal_node.line}" do
```
sds: This is incredibly hard to read. Breaking out the message into a separate variable so we can write:

```ruby
message = "..."
add_lint message do
  ...
end
```

...makes it much easier to read the block.

lib/scss_lint/linter/merge_equal_selector.rb
```diff
@@ -0,0 +1,60 @@
...
+        end
+      end
+
+      yield # Continue linting children
+    end
+
+    alias_method :visit_root, :check_node
+    alias_method :visit_rule, :check_node
+
+  protected
```
sds: `private` would be preferred here since we don't intend to have other instances of this class refer to its methods.

config/default.yml
```diff
@@ -92,9 +92,11 @@ linters:
     enabled: true
     style: exclude_zero # or 'include_zero'
 
-  MergeableSelector:
+  MergeEqualSelector:
```
sds: I think a better name in this case is `MergeEquivalentSelector`, even if perhaps more verbose.

lib/scss_lint/linter/nestable_selector.rb
```diff
@@ -0,0 +1,74 @@
+module SCSSLint
+  class Linter::NestableSelector < Linter
+    include LinterRegistry
+
+    def check_node(node)
+      visits = Set.new
+
+      node.children.each do |child_node|
+        nestable_node = find_nestable_node child_node, node.children
```
sds: We try to following the Jim Weirich philosophy which [includes parentheses when the return value matters](https://twitter.com/jimweirich/status/373814960788221952).

```ruby
nestable_node = find_nestable_node(child_node, node.children)
```

lib/scss_lint/linter/nestable_selector.rb
```diff
@@ -0,0 +1,74 @@
...
+      node.parsed_rules.members.count == 1
+    end
+
+    def nested?(node1, node2)
+      return false if node1 == node2
+      return false unless single_rule?(node1) && single_rule?(node2)
+
+      rule1 = node_rule(node1)
+      rule2 = node_rule(node2)
+      subrule?(rule1, rule2) || parentrule?(rule1, rule2)
```
sds: Can we call these `sub_rule?` and `parent_rule?` to separate the words?

sds:
Still some RuboCop errors and general style concerns that need to be addressed. Also, documentation needs to be added describing what these linters are intended to do.

sds:
Closing due to lack of response. Feel free to reopen with the issues identified corrected!

PR #536, opened 2015-07-28
Add GroupableSelector

brauliobo:
Depends on #534 

Scan for equal (with the same child prop names/values) selectors and suggest to group them.

### DIFF
--- a/lib/scss_lint/linter/groupable_selector.rb
+++ b/lib/scss_lint/linter/groupable_selector.rb
@@ -1,0 +1,64 @@
+# add missing comparison for literals
+module Sass::Script::Tree
+  class Literal
+
+    def == other
+      self.class == other.class and self.value == other.value
+    end
+  end
+end
+
+module SCSSLint
+  class Linter::GroupableSelector < Linter
+    include LinterRegistry
+
+    def check_node(node)
+      visits = Set.new
+      node.children.each do |child_node|
+        groupable_node = find_groupable_node child_node, node.children
+        next unless groupable_node
+        next if visits.include? [child_node, groupable_node] or
+                visits.include? [groupable_node, child_node]
+        visits << [child_node, groupable_node]
+
+        add_lint child_node.line, "Group identical rule `#{node_rule child_node}` on line #{child_node.line} " \
+        "with rule `#{node_rule groupable_node}` on line #{groupable_node.line}" do
+          node.children.delete groupable_node
+          child_node.parsed_rules = nil
+          child_node.rule.first << ", #{groupable_node.rule.join}"
+        end
+      end
+
+      yield # Continue linting children
+    end
+
+    alias_method :visit_root, :check_node
+    alias_method :visit_rule, :check_node
+
+  protected
+
+    # TODO: currently find only nodes on the same level, as node.rule
+    # does not know about its parent rules
+    def find_groupable_node(node, level_rule_nodes)
+      return unless node.is_a? Sass::Tree::RuleNode
+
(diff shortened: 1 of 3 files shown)


lib/scss_lint/linter/groupable_selector.rb
```diff
@@ -0,0 +1,64 @@
+# add missing comparison for literals
+module Sass::Script::Tree
+  class Literal
+
+    def == other
+      self.class == other.class and self.value == other.value
+    end
+  end
+end
```
sds: We should not be defining this override in the `GroupableSelector` linter. It's unorthodox and could change the behavior of other linters which rely on the current behavior of the `==` method (there may not be any right now, but that could change).

Can we have a helper method `same_literal?` defined on `GroupableSelector` that accomplishes the same effect?

lib/scss_lint/linter/groupable_selector.rb
```diff
@@ -0,0 +1,64 @@
...
+      visits = Set.new
+      node.children.each do |child_node|
+        groupable_node = find_groupable_node child_node, node.children
+        next unless groupable_node
+        next if visits.include? [child_node, groupable_node] or
+                visits.include? [groupable_node, child_node]
+        visits << [child_node, groupable_node]
+
+        add_lint child_node.line, "Group identical rule `#{node_rule child_node}` on line #{child_node.line} " \
+        "with rule `#{node_rule groupable_node}` on line #{groupable_node.line}" do
```
sds: Split out the message into a separate variable to make this more readable, e.g.

```ruby
message = "..."
add_lint message do
  ...
end
```

lib/scss_lint/linter/groupable_selector.rb
```diff
@@ -0,0 +1,64 @@
...
+          node.children.delete groupable_node
+          child_node.parsed_rules = nil
```
sds: Does this do what you think it does? There are multiple related fields defined on `RuleNode`: `rule`, `parsed_rules`, `resolved_rules`, etc. It makes me uncomfortable that we're relying on the behavior of this specific rule and hoping it will continue to work the same way in future.

lib/scss_lint/linter/groupable_selector.rb
```diff
@@ -0,0 +1,64 @@
...
+      return unless node.is_a? Sass::Tree::RuleNode
+
+      level_rule_nodes.find do |sibling_node|
+        next unless sibling_node.is_a? Sass::Tree::RuleNode
+        groupable?(node, sibling_node)
+      end
+    end
+
+    def groupable?(node1, node2)
+      node1 != node2 and node1.children.all? do |c1|
```
sds: Please prefer `&&` over `and`.

spec/scss_lint/linter/groupable_selector_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  context 'when different elements' do
+    let(:scss) { <<-SCSS }
+      p {
+        background: #000;
+        margin: 5px;
+      }
+      a {
+        background: #000;
+        margin: 5px;
+      }
```
sds: Are there other examples of situations where this linter is useful? If this is the only case, I'm not sure having this linter is really worth it, as the number of times where you have the exact same set of properties in both selectors is likely quite rare.

sds:
Thanks for the pull request, @brauliobo. There are still a number of RuboCop errors and general style concerns that need to be addressed.

sds:
Closing due to lack of response. Feel free to reopen with the recommended suggestions implemented!

PR #539, opened 2015-07-30
Allow control comments continued from a line with other nodes

### LINKED ISSUE #373: Disable linter fails after comment on previous line
I use a commenting system with numbers, so I don't have to put all of them inline, which can become really sloppy:

````scss
// 1. Some comment about my background
.foo {
  background: 0; // [1]
  border-radius: 0;
}
````

When using an inline scss-lint:disable comment, it breaks when it comes right after a line with this comment, like so:

````scss
// 1. Some comment about my background
.foo {
  background: none; // [1]
  // scss-lint:disable ImportantRule
  border-radius: 0 !important;
  // scss-lint:enable ImportantRule
}
````
It now reports an error on line 5 to not use `!important`. Removing the `// [1]` comment makes the inline disable statement work like it should. The same goes for when I wrap the whole selector in the toggling comments.

srawlins:
Fixes #373 

The new test cases cover what was not working before, and what works now.

### DIFF
--- a/lib/scss_lint/control_comment_processor.rb
+++ b/lib/scss_lint/control_comment_processor.rb
@@ -47,15 +47,7 @@ module SCSSLint
   private
 
     def extract_command(node)
-      comment =
-        case node
-        when Sass::Tree::CommentNode
-          node.value.first
-        when Sass::Tree::RuleNode
-          node.rule.select { |chunk| chunk.is_a?(String) }.join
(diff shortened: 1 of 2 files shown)


sds:
Thanks for the fix, @srawlins!

Merged in 4cd853748070c0ae63c844231cde15eb38d70bae.

Issue #541, opened 2015-07-31
undefined method `members' for nil:NilClass - NestingDepth error


viveleroi:
Exception: https://pste.me/#/4bijx

Our current config: https://pste.me/#/AVXTq

- SCSS-Lint version: 0.40.1
- Sass version: 3.4.16
- Ruby version: 2.0.0

The offending content, which when removed no longer triggers the issue:

    /* Key Frames */

    @include keyframes(slideInLeft) {
        0% {
            @include transform(translate(100%, 0));
            opacity: 0;
        }

        100% {
            @include transform(translate(0, 0));
            opacity: 1;
        }
    }

    @include keyframes(slideOutRight) {
        0% {
            @include transform(translate(0, 0));
            opacity: 1;
        }

        100% {
            @include transform(translate(150%, 0));
            opacity: 0;
        }
    }

    @include keyframes(hide) {
        0% {
            opacity: 1;
        }

        100% {
            opacity: 0;
        }
    }

It appears that disabling scss-lint via a comment doesn't work, so this is a blocking issue for us. I may need to try ignoring the entire file.

This seems to affect 0.40.0 as well.

sethlopezme:
I am experiencing this issue, as well.

Exception: https://pste.me/#/JZJGN

My config: https://pste.me/#/VnLM8

- SCSS-Lint version: 0.40.1
- Sass version: 3.4.15
- Ruby version: 2.0.0

The offending code is below. I tried and failed to isolate exactly what the problem code was here.

```scss
.row {
	@include flex-container;

	// modifiers
	&.spaced {
		@include flex-spaced;
	}

	&.spread {
		@include flex-spread;
	}

	&.align-left {
		@include flex-align-left;
	}

	&.align-right {
		@include flex-align-right
	}

	&.valign-top {
		@include flex-valign-top;
	}

	&.valign-bottom {
		@include flex-valign-bottom;
	}

	&.equal-heights {
		@include flex-equal-heights
	}
}

[class*="col"] {
	&.valign-top {
		@include valign-top;
	}

	&.valign-bottom {
		@include valign-bottom;
	}

	&.valign-middle {
		@include valign-middle;
	}
}

@for $i from 1 through $grid-columns {
	$_col-size : (100%/$grid-columns * $i);

	.col-#{$i} {
		@include grid-col($i);
	}

	.offset-#{$i} {
		margin-left: $_col-size;
	}
}

@each $name, $val in $breakpoints {
[...]

PR #542, opened 2015-08-01
NestingDepth linter not handling rules with no selectors

cih:
Hi Guys,

Noticed #541 so I've taken a look at this, when I wrote the initial implementation I didn't realise that `node.parsed_rules` could potentially return `nil`. If my understanding is correct a `RuleNode` can in some cases return `nil` when it contains no CSS selectors so I've gone for a somewhat quick and dirty null object, maybe an implicit `return` statement would do the job?

Hope this helps!

### DIFF
--- a/lib/scss_lint/linter/nesting_depth.rb
+++ b/lib/scss_lint/linter/nesting_depth.rb
@@ -30,6 +30,7 @@ module SCSSLint
 
     def ignore_selectors?(node)
       return unless config['ignore_parent_selectors']
+      return unless node.parsed_rules
 
       simple_selectors(node.parsed_rules).all? do |selector|
         IGNORED_SELECTORS.include?(selector.class)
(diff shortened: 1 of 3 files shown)


lib/scss_lint/linter/nesting_depth.rb
```diff
@@ -30,8 +32,7 @@ def visit_rule(node)
 
     def ignore_selectors?(node)
       return unless config['ignore_parent_selectors']
-
-      simple_selectors(node.parsed_rules).all? do |selector|
+      simple_selectors(node.parsed_rules || null_selector).all? do |selector|
```
sds: Unless we intend to use the `null_selector elsewhere, it seems simpler (and less code) to have an early return.

```ruby
return unless node.parsed_rules
```

sds:
Thanks for the fix, @cih!

I agree an early `return` would be preferred in this case. Otherwise this looks good!

cih:
No worries, cheers for the feedback @sds, updated.

PR #544, opened 2015-08-04
Fixes #518 by adding notes to linters/README.md

### LINKED ISSUE #518: If a linter is disabled by default, this is not documented anywhere
... except `config/default.yml`.

Should the disabled-by-default linters (like BemDepth) be noted in `lib/scss-lint/linter/README.md`?

srawlins:


lencioni:
Thanks!

lib/scss_lint/linter/README.md
```diff
@@ -1457,6 +1465,8 @@ Reports lines containing trailing whitespace.
 
 ## TrailingZero
 
+**Disabled by default**
```
sds: I'm actually surprised this is disabled by default. Unfortunately I didn't think to explain why I disabled it by default in the original commit 800a67bb.

PR #551, opened 2015-08-10
Add support for flexbox-related properties into SMACSS property sort order file

ingdir:
The existing SMACSS property sort file puts all flexbox-related properties to the end of the list (because they are not explicitly defined), which does not work well with modern CSS.

This PR addresses this inconsistency by adding a list of all common flexbox-related properties to the "Box" section. Old flexbox syntax is not considered as it has the same status as vendor-specific properties.

### DIFF
--- a/data/property-sort-orders/smacss.txt
+++ b/data/property-sort-orders/smacss.txt
@@ -10,6 +10,19 @@ right
 bottom
 left
 
+flex
+flex-basis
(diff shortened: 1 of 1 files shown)


sds:
Merged in 14417874f413d3ff0f2ebe9dfd1e2eb728d966ef. Thanks!

PR #554, opened 2015-08-11
Fix newlines in MergeableRule lint message

### LINKED ISSUE #545: MergableSelector output can't handle multiline selectors
When running the latest `scss-lint`, I noticed that some of the `MergeableSelector` warnings were being truncated like so:

    web/assets/sass/base.scss:100 [W] MergeableSelector: Merge rule `.dialog,

The root cause of this truncation was because I was piping the output through `grep` to see all of the `MergableSelector` warnings and the selector on line 100 had multiple lines. Without piping through grep, I get:

    web/assets/sass/base.scss:100 [W] MergeableSelector: Merge rule `.dialog,
    .modal` with rule on line 70

However, it seems to me like the newlines in the selector add no semantic value and instead make it slightly harder to parse (that is with your eyes, not with code :smile:) the list of errors.

In a sea of warning messages, I find this easier to read:

    web/assets/sass/base.scss:100 [W] MergeableSelector: Merge rule `.dialog, .modal` with rule on line 70

Could newlines in the output be stripped so that warnings for multiline selectors appear on one line?

srawlins:
Fixes #545 

Example `d.scss`:

```scss
.foo,
.bar {
  margin: 1px;
}

.foo,
.bar {
  padding: 1px;
}
```

Linting used to produce:

```
$ bundle exec bin/scss-lint d.scss
d.scss:6 [W] MergeableSelector: Merge rule `.foo,
.bar` with rule on line 1
```

Becomes:

```
$ bundle exec bin/scss-lint d.scss
d.scss:6 [W] MergeableSelector: Merge rule `.foo, .bar` with rule on line 1
```

### DIFF
--- a/lib/scss_lint/linter/mergeable_selector.rb
+++ b/lib/scss_lint/linter/mergeable_selector.rb
@@ -11,8 +11,10 @@ module SCSSLint
         seen_nodes << child_node
         next unless mergeable_node
 
+        rule_text = node_rule(child_node).gsub(/(\r?\n)+/, ' ')
+
         add_lint child_node.line,
-                 "Merge rule `#{node_rule(child_node)}` with rule " \
+                 "Merge rule `#{rule_text}` with rule " \
                  "on line #{mergeable_node.line}"
       end
 


srawlins:
Updated regex. Thanks.

sds:
Merged in 8c4d9464353f05adf1fa164edbe161c2468d3ad6. Thanks!

Issue #556, opened 2015-08-17
Linter Request: Discourage transition all


baileyparker:
I think the sentiment (especially on teams) regarding `transition: all` is to avoid it in most cases: http://stackoverflow.com/questions/8947441/css3-transitions-is-transition-all-slower-than-transition-x

Not so much from a performance perspective, of course, but I think most will recognize how it is possible that `transition: all` will unintentionally affect properties it was never meant to.

That is why I propose adding a linter that warns against uses of `all` in the `transition` or `transition-property` properties. If there's enough interest and someone is willing to merge this in, I can make a PR for this.

Anahkiasen:
:+1: I try to do this myself but one sometimes slip from me, would love a rule to enforce it.

sds:
Implemented in #556.

PR #557, opened 2015-08-17
Adds linter for transition all (#556)

baileyparker:
This PR addresses #556 by adding a `TransitionAll` linter that enforces explicit declaration of properties to be transitioned instead of using `transition: all` or `transition-property: all` (discussion about the validity of this preference can be found in the issue).

I have set the linter to be disabled by default, because I wasn't sure if it should be enabled. I think it's a fairly common sense linter to enable, but if there is a compelling reason why it shouldn't be, it can be left as is.

I've updated the docs and tried to add as helpful of error messages as possible. The best I could come up with was:

> {transition/transition-property} should contain explicit properties instead of using the keyword all

I'm definitely open to alternative phrasings, though.

Also one thing to note: To test if the `all` keyword is used, I've done a word-boundaried regex on the value of `transition` or `transition-property`. This of course could report false positives in very strange cases, but it seems like this approach is used by other linters, so I assumed it was accepted. If there is a better way though, I'd love to hear about it.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -201,6 +201,9 @@ linters:
   TrailingZero:
     enabled: false
 
+  TransitionAll:
+    enabled: false
+
   UnnecessaryMantissa:
     enabled: true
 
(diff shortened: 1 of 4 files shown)


sds:
Thanks! Merged in 62641033111fb3f2dc08cc6a1c0c0fc5237a1e41.

PR #563, opened 2015-08-25
Adjust some operations for selector interpolation

### LINKED ISSUE #560: SpaceAroundOperator linter always reports operators in string interpolation
```
$ scss-lint --version
scss-lint 0.41.0
$ cat .scss-lint.yml
linters:
  SpaceAroundOperator:
    enabled: true
    style: one_space
$ cat test.scss
@for $i from 1 through 25 {
  .progress-bar[aria-valuenow="#{$i / 25 * 100}"] {
    opacity: $i / 25 + .03;
  }
}
$ scss-lint test.scss
test.scss:2 [W] SpaceAroundOperator: ` / 25 *` should be written with a single space on each side of the operator: ` / 25  *`
test.scss:2 [W] SpaceAroundOperator: ` / 25 * 100}"` should be written with a single space on each side of the operator: ` / 25 * 10 0}"`
```

Is it a bug?

srawlins:
Fixes #560 

I need to open an issue with Sass too, but probably hard to write a test case not involving scss-lint... :frowning: .

This refactoring involves an inner class, and opening up the protected `:source_from_range`. Let me know if anything's not kosher.

### DIFF
--- a/lib/scss_lint/linter/space_around_operator.rb
+++ b/lib/scss_lint/linter/space_around_operator.rb
@@ -3,20 +3,32 @@ module SCSSLint
   class Linter::SpaceAroundOperator < Linter
     include LinterRegistry
 
-    def visit_script_operation(node) # rubocop:disable Metrics/AbcSize
-      source = normalize_source(source_from_range(node.source_range))
-      left_range = node.operand1.source_range
-      right_range = node.operand2.source_range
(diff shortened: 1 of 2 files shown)


sds:
Thanks for the fix, @srawlins!

Exposing `source_from_range` is a bit unorthodox but given this is isolated to a single class I'm fine with it.

Issue #488, opened 2015-06-09
ColorVariable should flag uses of rgb/rgba/hsl/hsla


cH40z-Lord:
Today I've discovered a strange inconsistency between color definitions. SCSS-Lint is throwing a warning whenever you try to use a color definition like `#fff` outside of a SCSS variable. However, this is not true for colors defined via `rgb()` and `rgba()`, which feels kinda strange and not what I expect.

```css
.foo {
  // warning: foo.scss:2 [W] ColorVariable: Color literals like `#fff` should only be used in variable declarations; they should be referred to via variable everywhere else.
  backgorund-color: #fff;
}

.bar {
  // no warning/error
  background-color: rgba(255, 255, 255, .5);
  //background-color: rgb(255, 255, 255);
}
```

I'd love to have an option for this as well so we can improve our SCSS code even more. What do you think?

apfelbox:
Maybe regression by #402?

Here is my understanding / proposal of the `ColorVariable` linter:

| Type                                            | Code Examples                    | SHOULD report? | DOES CURRENTLY report? |
| ----------------------------------------------- | -------------------------------- | -------------- | ---------------------- |
| **Regular color definition**                    | `color: red;`                    | yes            | yes ✓                  |
|                                                 | `color: #fff;`                   | yes            | yes ✓                  |
|                                                 | `color: #ffffff;`                | yes            | yes ✓                  |
|                                                 | `color: rgb(0, 100, 200);`       | yes            | **no ✗**               |
|                                                 | `color: rgb(0%, 50%, 80%);`      | yes            | **no ✗**               |
|                                                 | `color: rgba(0, 0, 0, .5);`      | yes            | **no ✗**               |
|                                                 | `color: hsl(0, 100%, 50%);`      | yes            | **no ✗**               |
|                                                 | `color: hsla(0, 100%, 50%, .5);` | yes            | **no ✗**               |
| **Special CSS keywords**                        | `color: currentColor;`           | no             | no ✓                   |
|                                                 | `color: inherit;`                | no             | no ✓                   |
| **Resetting a color**                           | `color: transparent;`            | no             | no ✓                   |
| **Directly using a variable**                   | `color: $red;`                   | no             | no ✓                   |
| **Transforming a variable value in a function** | `color: rgba($red, .5);`         | no             | no ✓                   |
|                                                 | `color: lighten($red, 5%);`      | no             | no ✓                   |


Tested with scss-lint 0.39.0 and sass 3.4.14

[...]

PR #569, opened 2015-08-28
Improve ColorVariable to flag uses of rgb/rgba/hsl/hsla

cih:
Hey Folks, 

I've started working on https://github.com/brigade/scss-lint/issues/488 and I was hoping to get some feedback before completing this. I've added some specs based on the examples in the matrix and these are now passing. There are a couple of things to note, as we're simply checking all of the color function arguments are literals then we won't raise a lint in the case where we have a mixture of variables and literals, so the example below won't raise a lint, would my implementation need to handle this too?

```css
  a {
    color: rgb($some-var, 50%, 80%);
  }
```

Also I've changed this [spec](https://github.com/cih/scss-lint/blob/1759e748bc4b0017c94154561c56baf773e04202/spec/scss_lint/linter/color_variable_spec.rb#L123-L131) to expect a lint to be reported, this was previously expected to not report, what would we want to happen in this case. Any feedback would be appreciated.

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Fix `SpaceAroundOperator` linter to not report false positives operators in
   interpolation
+* Improve `ColorVariable` to flag uses of rgb/rgba/hsl/hsla
 
 ## 0.41.0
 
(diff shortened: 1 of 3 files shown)


sds:
This looks good to me, @cih. Thanks for the PR!

PR #573, opened 2015-08-31
Support any command line flags in the rake task

### LINKED ISSUE #570: Feature Request: Ability to specify formatter via Rake task or config file
For our setup, we'd like to use the checkstyle formatter in our CI environment. The `RakeTask` class doesn't allow anything more than a set of files to scan and a config file. As a workaround, we are invoking `scss-lint` via a `sh` command:

```ruby
    task :scss_lint do
      sh "scss-lint " \
         "--require 'scss_lint_reporter_checkstyle' " \
         "--format 'Checkstyle' " \
         "--out '#{scss_lint_file}'"
    end
```

srawlins:
Fixes #570

This PR allows blanket option passing into the Rake task (similar to other tools' implementations, like [Cucumber](https://github.com/cucumber/cucumber/wiki/Using-Rake)).

I'd like to somehow deprecate the `@config` piece, so that all options are just passed in with `@args`, and display that it is deprecated, but I didn't implement this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 [![Dependency Status](https://gemnasium.com/brigade/scss-lint.svg)](https://gemnasium.com/brigade/scss-lint)
 
 `scss-lint` is a tool to help keep your [SCSS](http://sass-lang.com) files
-clean and readable. You can run it manually from the command-line, or integrate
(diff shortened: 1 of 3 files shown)


sds:
Thanks for the pull request, @srawlins.

Merged in e2a96cb.

PR #575, opened 2015-09-01
Include MIT-LICENSE in gemspec

bkeepers:
The MIT License should be included with any distribution of the code. This adds `MIT-LICENSE` to the gemspec.


sds:
Thanks! 

Merged in 6eaf4ed454814afe94ddf275de06a293e8551fac.

PR #577, opened 2015-09-01
Only initialize, and call #run_linter for enabled linters

srawlins:


sds:
Thanks! 

Merged in 41f3750f67135e476a79493de15c3dfe7edc41a5.

PR #580, opened 2015-09-02
Use Regexp less. Just for little performance gains

srawlins:


sds:
Thanks for your work on this @srawlins.

I'm not sure how significant these sorts of optimizations are, but none of these changes hurt readability so they seem fine.

Merged in 954e22fd2f9ab352d07953ec6c62836abd747d6e.

PR #584, opened 2015-09-02
Conservatively process control commands

srawlins:
Processing control command comments on every single node in a file (`Linter#visit`), for every single linter, seems heavy. I've changed the workflow to only search for control commands if the file has any at all (presumably, most files don't).

Linting Bootstrap 4 Alpha took ~11 seconds before this fix, and ~9 seconds after this fix.


sds:
Nice find, @srawlins. Thanks again for your efforts on improving the tool's performance.

Merged in 33cf2bcd615d37d1797ea3fc4dff4a50a0506576.

PR #590, opened 2015-09-09
Fix VariableForProperty lint when used with !important

ivantsepp:
For this example: 
```scss
.test {
  color: $silver !important;
}
```
The linter outputs:
```
test.scss:2 [W] ImportantRule: !important should not be used
test.scss:2 [W] VariableForProperty: Property color should use a variable rather than a literal value
```
I added a PR to handle the case when a variable is used with `!important`. 

---

Actually, when I looked at this lint a little more closely, I'm curious how this rule would play with something like `font: $large arial, sans-serif;`. What is the expected output?

### DIFF
--- a/lib/scss_lint/linter/variable_for_property.rb
+++ b/lib/scss_lint/linter/variable_for_property.rb
@@ -15,12 +15,20 @@ module SCSSLint
       return unless @properties.include?(property_name)
       return if ignored_value?(node.value)
       return if node.children.first.is_a?(Sass::Script::Tree::Variable)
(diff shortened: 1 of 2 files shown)


sds:
Thanks for the pull request, @ivantsepp!

Merged in f25e03978be633e329891db3e518d2627e4ac336.

I would say that examples like `font: $large arial, sans-serif;` are a violation of what `VariableForProperty` is trying to enforce. If you'd be interested in submitting a PR addressing that case, we'd be happy to merge it!

Issue #602, opened 2015-09-25
TrailingSemicolon with !default, !important


mrowles:
I am getting errors with my SCSS where it thinks that there is no semicolon if I have a `!important` flag after the value.

    Case 1:
    $a-var-height: 100 !important; // warning here
    $b-var-height: 100px !important; // warning here
    
    Case 2:
    $b-var-height: 100;
    .b-var {
        height: em($b-var-height) !important; // warning here
    }

    Case 3:
    $c-var-height: 100px;
    .c-var-height {
        height: $c-var-height !important; // warning here
    }

So as you can see, I have covered all bases to ensure that at least somewhere I'd be writing the correct code. I have the rule in my .scss-lint.yml to allow for important rules:

    ImportantRule:
        enabled: true

My versions are:

* Sass 3.4.18
* SCSS Lint 0.42.0
* Ruby 2.0.0p481

Any help would be appreciated.

mlarcher:
I just stumbled on the same issue, it seems the TrailingSemicolon rule doesn't account for this case.

samuelmartineau:
:+1: 

Same there with **!global;**

Sass 3.4.18
SCSS Lint 0.42.0
Ruby 1.9.3p484 (2013-11-22 revision 43786) [x86_64-linux

lencioni:
Thanks for the report. I wasn't able to reproduce this problem with the cases you specified, but I was able to reproduce a problem when using `!default` and `!global` flags on variable declarations. I will push a fix for that shortly.

I may be mistaken, but I believe that `ImportantRule` will warn on any uses of `!important`. If you want to use `!important` then you should disable this rule. What output are you seeing when running scss-lint on the repro cases you posted?

XhmikosR:
Hi, @lencioni. See https://travis-ci.org/twbs/bootstrap/builds/82182873

lencioni:
@XhmikosR thanks for the link! It would be helpful to be able to see the source that it is breaking on here, however, I'm having a hard time finding the .scss files that are reported there (`scss/_variables.scss` and `scss/bootstrap-grid.scss`). Can you point me in the right direction?

XhmikosR:
@lencioni: it's in our v4-dev branch; that's where we switched to Sass. Or check the branch I pushed the scss lint update, it's based on the same one.

lencioni:
@XhmikosR great, thanks! I'm pretty sure that the fix I just pushed will fix this problem for you.

[...]

PR #605, opened 2015-09-25
Fix TrailingSemicolon for variables with !default and !global

lencioni:
The TrailingSemicolon rule was not checking guarded or global variable
declarations properly. I determined that this is because we are using
`node.expr` for the source range, which in cases like:

  $foo: bar !default;

gives us a range that looks like:

  bar

which causes the linter to fail. Instead, if we just use the `node`
source range instead, we will get

  foo: bar !default;

so our checks work as expected. I considered always using `node` source
range, but I figured that if it is working for other cases that it would
be better to leave it as is.

Addresses #602

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # SCSS-Lint Changelog
 
+## master (unreleased)
+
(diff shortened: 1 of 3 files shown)


sds:
Nice :+1:

PR #606, opened 2015-09-25
Check !default, !global, and !optional flags in BangFormat

lencioni:
This linter was not visiting variable declarations or `@extend`
directives, so it was not checking for `!default`, `!global`, and
`!optional`.


sds:
Awesome, looks good :+1:

PR #607, opened 2015-09-25
Fix newlines around operators

### LINKED ISSUE #598: SpaceAroundOperator should allow newlines as well
Right now SpaceAroundOperator has only two `style` config choices:
`one_space` or `no_space`. How about a `one_space_or_newline` like SpaceAfterPropertyColon has?

This would support the uses of string concatenation in order to keep a maximum width on files, like:

```scss
p:before {
  content: 'Something crazy long and weird and longer than '
      + '80 characters.';
}
```

I think the allowed whitespace of `one_space_or_newline` would allow:

* one space character on each side of an operator, or
* one space character on one side of the operator and a newline followed by any number of space characters on the other side (indentation is not measured).

```scss
a {
  content: 'hello ' + 'world';
  font-family: 'hello ' +
      'world';
  something-else: 'hello '
      + 'world';
}
```

(all legal SCSS.)

So, not allowed:

* an operator on a line by itself
* multiple newlines
* tab characters

What do you think? Happy to do the work.

srawlins:
Fixes #598 by forgiving space around an operator with a newline.

### DIFF
--- a/lib/scss_lint/linter/space_around_operator.rb
+++ b/lib/scss_lint/linter/space_around_operator.rb
@@ -21,6 +21,7 @@ module SCSSLint
       yield
     end
 
+    # Making a public version of #source_from_range, a private method.
     def source_fm_range(range)
       source_from_range(range)
     end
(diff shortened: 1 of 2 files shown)


sds:
Thanks @srawlins!

PR #610, opened 2015-09-28
Don't lint HexValidation on mid-word hex strings

### LINKED ISSUE #597: HexValidation incorrectly validates on strings
```shell
$ cat hex.scss
polyfill-unscoped-rule {
  content: 'div#ab-overlay-container';
}
$ bundle exec bin/scss-lint hex.scss
hex.scss:2 [W] HexValidation: Colors must have either three or six digits: `#ab`
```

HexValidation should not lint on quoted strings, ever, right?

srawlins:
Fixes #597 

I went with "non-word" (`\W`) separations, because word-break separations (`\b`) don't work with the `#` character.

### DIFF
--- a/lib/scss_lint/linter/hex_validation.rb
+++ b/lib/scss_lint/linter/hex_validation.rb
@@ -6,7 +6,7 @@ module SCSSLint
     def visit_script_string(node)
       return unless node.type == :identifier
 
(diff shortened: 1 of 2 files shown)


sds:
Merged in 0a9f0245294eb94c00a4d963c871d06b32b9be0f. Thanks!

PR #612, opened 2015-10-01
Fix `ColorVariable` to ignore color functions in variable declaration

cih:
Hey Folks,

Not sure what way you are wanting to go on the discussion in https://github.com/brigade/scss-lint/issues/488 but this change updates the `ColorVariable` linter to ignore color functions within variable declarations. If there are any other cases that might need to be handled then let me know and I can update it.

Cheers!


sds:
Thanks for the fix, @cih!

PR #622, opened 2015-10-23
Simplify Indentation linter; add memoization

srawlins:
I was looking into Indentation (as the slowest linter; this PR doesn't really fix it), and fixed the following:

* initialize all config variables in `visit_root`
* don't pass those config variables around
* combine all lint messages into `lint_message`
* move the root ruleset checking into its own method; removes some rubocop disables
* overall fewer lines


sds:
Merged in 1652bfea532cdeba3f5a91b5a71dd994916c6676. Thanks!

PR #623, opened 2015-10-26
Add a @@simple_name to Linter

srawlins:
This is a performance boost. I'm adding a `@@linter_name` to each Linter class as it inherits `Linter`. I used stackprof to profile linting (4 copies of) Bootstrap 4's SCSS files, and found that a lot of time was spent in Config.linter_name, Linter#name and a **ton** of time in SelectorVisitor#selector_node_name.

This refactoring computes the simple name of a linter once, memoizing the result for all of these methods.

In that Bootstrap 4 case, this drove the time to lint down from 20 seconds to 15 seconds!

### DIFF
--- a/lib/scss_lint/cli.rb
+++ b/lib/scss_lint/cli.rb
@@ -204,7 +204,7 @@ module SCSSLint
       puts 'Installed linters:'
 
       linter_names = LinterRegistry.linters.map do |linter|
-        linter.name.split('::').last
+        linter.simple_name
       end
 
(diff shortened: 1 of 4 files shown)


sds:
Merged in 9ecc9aee3ee23119ef8ce4c12af707ffd02e0c30 with some amendments in d117ef4e1907a592e889f37f2176ed1da9780ab4 to fix RuboCop warnings.

Thanks!

PR #624, opened 2015-10-26
Improve character_at with other methods

srawlins:
This is a performance boosting change.

* move `character_at` to be private.
* Add an `offset_to` method to be used instead of `character_at` in a few cases. These were cases where a method was walking character by character, searching for something specific. This is made much faster with `offset_to`.


sds:
Merged in 8af3c89. Thanks!

PR #631, opened 2015-11-07
Add a whitelist parameter to MergeableSelector

### LINKED ISSUE #540: MergeableSelector should allow a whitelist
For anyone using Polymer (and some versions of Angular), there are a few bizarre CSS rules that you can sprinkle into your CSS that do magical things [1] [2]. Here's an example from Polymer's docs:

```css
polyfill-next-selector { content: ':host > *' }
::content > * { }

polyfill-rule {
  content: '.bar';
  background: red;
}

polyfill-rule {
  content: ':host.foo .bar';
  background: blue;
}
```

I know, it's bizarre. Because of how these "rules" work, you will repeat them multiple times throughout your CSS, and MergeableSelector will report on them. Can MergeableSelector support a whitelist of selectors that the developer intends to repeat, because a JS library does magic with it? I would use it like:

```yaml
MergeableSelector:
  whitelist: ['polyfill-next-selector', 'polyfill-rule', 'polyfill-unscoped-rule']
```

[1] https://www.polymer-project.org/0.5/docs/polymer/styling.html#at-polyfill
[2] https://github.com/angular/angular.dart/wiki/CSS-Shim

srawlins:
Fixes #540.

### DIFF
--- a/lib/scss_lint/linter/README.md
+++ b/lib/scss_lint/linter/README.md
@@ -685,6 +685,9 @@ h1 {
 Configuration Option | Description
 ---------------------|------------------------------------------------------------------
 `force_nesting`      | Ensure rule sets which can be nested are nested (default **true**)
+`whitelist`          | A list of selectors that can MergeableSelector, list those used in [CSS Shims][]
+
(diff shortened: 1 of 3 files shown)


lencioni:
Thanks again for the code and tests!

Issue #538, opened 2015-07-29
EmptyLineBetweenBlocks should lint around @if, etc.


srawlins:
It seems like EmptyLineBetweenBlocks should lint around `@if`, no? It currently doesn't:

```bash
$ cat -n a.scss
     1  a {
     2    margin: 1px;
     3    @if (1 + 1 == 2) {
     4      padding: 2px;
     5    }
     6    .b {
     7      margin: 1px;
     8    }
     9  }
$ bundle exec ./bin/scss-lint a.scss
a.scss:6 [W] EmptyLineBetweenBlocks: Rule declaration should be preceded by an empty line
$
```

I expected an additional lint on line 3.

I haven't checked other block types.

sds:
`@if` is tricky because of `@else if`/`@else`. I also think that since it deals with control flow the same rules need not necessarily apply.

I would certainly be open to extending `EmptyLineBetweenBlocks` to accept a configuration option that allows you to choose which type of block statements for which you want to enforce blank lines.

srawlins:
For reference, `EmptyLineBetweenBlocks` currently appears to expect spaces around:

* `@mixin` declarations
* `@include` declarations
* rule declarations
* `@function` declarations

and _not_ expect spaces around:

* `@media` declarations
* `@if` declarations
* `@at-root` declarations
* `@for` declarations
* `@each` declarations
* `@while` declarations

I would argue that `EmptyLineBetweenBlocks` should lint on `@media` and `@at-root` out of the box. But maybe control flow directives are special. So the configuration option `special_block_types` (?) could take the value `:control_directives`. I can't really identify better categorizations...

sds:
I think allowing you to name each type of node separately would be sufficient, rather than introducing fancy names for entire types of blocks. So adding a `spaced_blocks` option where you can specify:

```yaml
spaced_blocks:
  - media
  - if
  - for
  - each
  ...
```

...or similar.

PR #632, opened 2015-11-08
EmptyLineBetweenBlocks lints on @media and @at-root now

srawlins:
This partially addresses #538.

This only affects @media and @at-root, which are not control flow. It seems like EmptyLineBetweenBlocks should always expect empty lines around these.


lencioni:
:+1: Thanks for the code and tests!

PR #635, opened 2015-11-12
Blacklist properties

dspautz:
Please review my additional code to to support blacklisted properties for PropertySpelling.
Let me know if something is wrong.

Thank you and have a nice day


lib/scss_lint/linter/property_spelling.rb
```diff
@@ -9,7 +9,9 @@ class Linter::PropertySpelling < Linter
                            .to_set
 
     def visit_root(_node)
-      @extra_properties = config['extra_properties'].to_set
+      @extra_properties = config['extra_properties'] ? config['extra_properties'].to_set : []
+      @disabled_properties = config['disabled_properties'] ? config['disabled_properties'].to_set : []
```
sds: Can write as:

```ruby
@extra_properties = Array(config['extra_properties']).to_set
@disabled_properties = Array(config['disabled_properties']).to_set
```

.scss-lint.yml
```diff
@@ -0,0 +1,213 @@
...
+    properties: [background-color, border-color, color]
+
+  VendorPrefix:
+    enabled: false
+
+  ZeroUnit:
+    enabled: true
+
+  Compass::*:
+    enabled: true
```
sds: This file should be removed.

spec/scss_lint/linter/property_spelling_spec.rb
```diff
@@ -55,6 +55,31 @@
     end
   end
 
+  context 'when disabled properties are specified' do
+    let(:linter_config) { { 'disabled_properties' => ['made-up-property'] } }
+
+
```
sds: Lose the extra line

sds:
Thanks for the pull request, @dspautz.

It seems like the purpose of this change is to report blacklisted properties separately. Right now a blacklisted property will report "Unknown property ...". That doesn't seem correct. Instead, we should report a different message in the case the property is blacklisted, e.g. "Property '_property_' is prohibited."

dspautz:
Please checkout my changes. Let me know when I can help.

sds:
Thanks for address those comments, @dspautz.

Merged in f165d94.

Issue #364, opened 2015-02-14
Color not passed when running through Ruby


fj:
If you run `scss-lint` directly, the output is colored:

```
scss-lint path/to/scss
<colorized output>
```

But if you run it through Ruby, it isn't:

```
ruby -e "puts %x{scss-lint path/to/scss}"
<uncolorized output>
```

The expected result is that the output is colored. Is it possible to follow RSpec's lead here and add a `--color` switch that always passes color?

sds:
I wouldn't go so far to say the _expected_ behavior is for it to output color. Most CLI tools switch to non-colored output when outputting to a non-TTY (`grep`, `ack`, to name a few).

I would say that it's reasonable to expect `scss-lint` to provide `--color`/`--no-color` flags to force either.

This will require some refactoring to extract all the random `puts` calls into an output class of some kind so the behavior can be toggled in a single place instead of throughout the code.

A pull request would be welcome if you want this sooner rather than later.

fj:
It's not urgent, but I'd be happy to take a look. Is that how `scss-lint` generally writes things to `$stdout`, via `puts`?

sds:
Yes, `scss-lint` currently uses `puts` to display any output. 

See the [Logger](https://github.com/causes/haml-lint/blob/master/lib/haml_lint/logger.rb) class of the `haml-lint` project for a typical approach to this problem.

Otherwise I'll get to it when I can.

sds:
Implemented in #651.

PR #651, opened 2015-12-10
Add --color and --no-color options

lencioni:
We want to allow people to force the use or non-use of color when
running scss-lint. This commit does this by extracting the logging into
a Logger class and adding `--color` and `--no-color` options.

For this code, I essentially brought over the same stuff that was done
in HamlLint.

Addresses #364


lencioni:
Ack, it looks like this might not entirely do the job, since default reporter still adds color via the rainbow gem. Since the reporters in this project return strings instead of printing them, and since Logger prints things, how do you think we should tackle this @sds?

lencioni:
I think I'm just going to add a logger argument to the `Reporter` initialize and expose some methods for coloring text on the logger.

sds:
Thanks for the pull request, @lencioni.

Yes, we'll need to rewrite the reporters to use the `Logger` class, as all output should be delegated to it.

lencioni:
Alright, I just pushed a new version. Let me know what you think!

sds:
Thanks so much for implementing this, @lencioni!

lencioni:
\o/

PR #658, opened 2015-12-29
Add background-attachment to SMACSS property sort order

bkzl:
Nothing too special just added a `background-attachment` to SMACSS list of properties for a `property_sort_order` linter.


lencioni:
Seems reasonable. :+1:

PR #661, opened 2015-12-30
Add CSS Logical Properties to properties data

### LINKED ISSUE #660: Support for CSS Logical Properties
Is it possible to add the support for the CSS Logical Properties?

http://caniuse.com/#search=logical%20properties

For example
```
[W] PropertySpelling: Unknown property padding-inline-end
[W] PropertySpelling: Unknown property padding-inline-start
```

Thank you

lencioni:
This will make it so that rules such as PropertySpelling will allow
these valid CSS properties.

I pulled the list from this MDN article:

  https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties

Fixes #660.


sds:
Thanks!

Merged in f555d9d5b54f0afd018f8b14bc9fb5b48efa8b5f.

PR #664, opened 2015-12-31
Added initial to list of ignored for VariableForProperty;

abea:
Added `initial` to list of ignored values for VariableForProperty as it is a CSS keyword in the vein of `inherit` and `transparent`, which are already ignored.


abea:
Closing to update test.

PR #665, opened 2015-12-31
Add initial varforprop

abea:
Added `initial` to list of ignored values for VariableForProperty as it is a CSS keyword in the vein of `inherit` and `transparent`, which are already ignored.


lencioni:
:+1:

sds:
Thanks!

Merged in bb92fdd2d8fa46eb9bb2ec4a1655888a689ce355.

PR #667, opened 2016-01-05
Add Webkit-specific pseudo elements to PseudoElement

thomasthesecond:
There appears to be two additional pseudo elements for Webkit-based browsers.

- `::-webkit-search-decoration`
- `::-webkit-search-results-decoration`

[Reference](https://davidwalsh.name/remove-search-box-buttons-webkit)


thomasthesecond:
Good find! I'm happy to add these in as well.

sds:
Thanks! 

Merged in 2b6063a8e3cbe2fa9a0f4788b7ae2687a64f3550.

PR #668, opened 2016-01-05
Add Ruby 2.3 to Travis CI build matrix

lencioni:
Ruby 2.3.0 was recently released and we want to make sure that our tests
are working under this version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ rvm:
   - 2.0.0
   - 2.1
   - 2.2
+  - 2.3
   - jruby-19mode
 
 before_script:


lencioni:
Looks like this won't pass until https://github.com/travis-ci/travis-ci/issues/5361 is resolved. In the meantime we could use 2.3.0 instead of 2.3, but I think we may as well just wait.

sds:
Thanks @lencioni!

Travis has been really slow to fix this issue. I've modified your PR to use the explicit "2.3.0" for now since it's better than nothing.

Merged in 47d3dfcb21f30509e6660ef6ba8acaa430089ff0.

sds:
Had to follow up with bb69ac673903ca1de4a3c2b679cd4deea05c8f16 in order to get the build working.

PR #669, opened 2016-01-08
adding ''backdrop-filter" to allowed properties

erwinsmit:
See https://developer.mozilla.org/en-US/docs/Web/CSS/backdrop-filter


sds:
Thanks!

Merged in 5f813e8c23a9c7c4c84341773b2c09400e7d96ce.

PR #673, opened 2016-01-10
Add -webkit-search-decoration to list of pseudo elements

paulrobertlloyd:
The absence of this element was showing a linting error, as it appears in normalise.css (https://github.com/necolas/normalize.css/blob/master/normalize.css#L367)


paulrobertlloyd:
Oops, this is a dupe of https://github.com/brigade/scss-lint/pull/667

PR #679, opened 2016-01-18
SMACSS sort order for transition and transform

tobiashm:
Added missing sort order entries for indivdual properties for transition and transform.


sds:
Thanks!

Merged in a0465589d15ac35ec54d340e40642e898cd50d40.

PR #691, opened 2016-01-27
TrailingSemicolon: Allow single-line nested ListLiterals and MapLiterals

### LINKED ISSUE #690: TrailingSemicolon: Incorrect results for single line nested ListLiterals and MapLiterals
In this code, the TrailingSemicolon linter incorrectly reports missing semicolons:
```scss
    $foo: ((1, 2));
    $bar: ((1 2));
    $baz: ("a": ("b", "c"));
    $baz: ("a": ("b" "c"));
```

---

I'm pretty sure this is caused by [trailing_semicolon.rb:24](https://github.com/brigade/scss-lint/blob/8859b8a1fcc4d8ed810ec1f79bea75afc40b993d/lib/scss_lint/linter/trailing_semicolon.rb#L24), and removing the line (and the `&&`) appears to solve all problems.
```ruby
      if (node.expr.is_a?(Sass::Script::Tree::ListLiteral) ||
          node.expr.is_a?(Sass::Script::Tree::MapLiteral)) &&
         !node_on_single_line?(node) # <-- here
        return check_semicolon(node)
      end
```

Am I missing something? I really don't understand why that check has been introduced.

ThomasR:
Also, add a ton of test cases for lists.
This fixes #690.

### DIFF
--- a/lib/scss_lint/linter/trailing_semicolon.rb
+++ b/lib/scss_lint/linter/trailing_semicolon.rb
@@ -19,9 +19,8 @@ module SCSSLint
       # `node.expr` will give us everything except the last right paren, and
       # the semicolon if it exists. In these cases, use the source range of
       # `node` as above.
-      if (node.expr.is_a?(Sass::Script::Tree::ListLiteral) ||
-          node.expr.is_a?(Sass::Script::Tree::MapLiteral)) &&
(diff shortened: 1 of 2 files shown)


sds:
Merged in 20523e4.

Issue #655, opened 2015-12-17
Feature request: More informative output


adsteel:
Currently the output of running `$ scss-lint` is just the error lines. No error count. No message if there are no errors. It would be nice to have:

- An error count if there are errors
- A success message of there are no errors
- An error raise if there are errors, to trigger a failed build on CI tools like CircleCI or TravisCI

sds:
Hey @adsteel, thanks for suggesting ways we can improve the tool.

I'll answer each of your points separately

>- An error count if there are errors

Error count is easily obtainable by counting the lines of output with the default formatter, e.g. `scss-lint ... | wc -l`.

>- A success message of there are no errors

This default behavior is the standard for most tools. If there is nothing interesting to report, `scss-lint` should say nothing interesting.

>- An error raise if there are errors, to trigger a failed build on CI tools like CircleCI or TravisCI

The tool already returns a non-zero exit status if errors are reported (see the [exit codes documentation](https://github.com/brigade/scss-lint#exit-status-codes)), and thus will trigger failed builds in CI.

This default behavior is valuable for integration with other tools. I am completely open to accepting a pull request adding a new formatter that adds the additional information you request, but I don't see it as a common use case for the tool, especially if you're using it the way it was intended, for example via [Overcommit git hooks](https://github.com/brigade/overcommit) or in CI where all you care about is pass/fail status.

Thanks!

adsteel:
@sds Thanks for the quick response. I built a script to do all of that, but I was expecting a more informative output out of the package a la Rubocop, ESlint or RSpec. I wouldn't say that the current behavior is standard for most tools, given the two primary linters used have informative output. I may open a PR for that, but I may not and it seems like a worthwhile feature to leave an issue open for.

I also had to explicitly program the exit code in my script, otherwise the build passed on CircleCI, so whatever is happening inside scss-lint isn't bubbling up and actually triggering a failed build.

sds:
Ok, I'm fine to leave this open.

Regarding:
>I also had to explicitly program the exit code in my script, otherwise the build passed on CircleCI, so whatever is happening inside scss-lint isn't bubbling up and actually triggering a failed build.

[...]

PR #695, opened 2016-01-29
[feature] more informative test output

adsteel:
Addresses issue https://github.com/brigade/scss-lint/issues/655

scss_output

### DIFF
--- a/lib/scss_lint/runner.rb
+++ b/lib/scss_lint/runner.rb
@@ -2,13 +2,14 @@ module SCSSLint
   # Finds and aggregates all lints found by running the registered linters
   # against a set of SCSS files.
   class Runner
-    attr_reader :lints, :files
+    attr_reader :lints, :files, :lints_run_count
 
     # @param config [Config]
     def initialize(config, logger)
       @config  = config
       @lints   = []
       @logger = logger
+      @lints_run_count = 0
       @linters = LinterRegistry.linters.select { |linter| @config.linter_enabled?(linter) }
       @linters.map!(&:new)
     end
@@ -16,9 +17,10 @@ module SCSSLint
     # @param files [Array<Hash>] list of file object/path hashes
     def run(files)
       @files = files
-      @files.each do |file|
+      result = @files.each do |file|
         find_lints(file)
       end
+      result
     end
 
   private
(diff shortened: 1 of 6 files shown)


adsteel:
Outdated commit with original conversation [is here](https://github.com/adsteel/scss-lint/commit/da3ee26fe2e9f5926ef230ddf3cb55a7aaafccbd)

connorshea:
Would really like this merged, one major thing I feel is missing from the gem right now :)

sds:
Having a closer look at this I'm afraid this can't be merged as-is.

This results in this output being displayed for all different formatters, not just the default formatter. Thus it will break other formatters such as the JSON formatter since you're appending non-JSON to the end of the output.

You'll need to do some refactoring to have this happen in the `DefaultFormatter` class. I'm definitely in support of making SCSS-Lint's default output better, but we can't be breaking other output formats in the process. It needs to be isolated to the `DefaultFormatter`.

lencioni:
Yeah, that's what I meant by https://github.com/adsteel/scss-lint/commit/da3ee26fe2e9f5926ef230ddf3cb55a7aaafccbd#commitcomment-15776473

adsteel:
@sds Just getting back to this now, but I can't find a `DefaultFormatter` class.

adsteel:
@sds @lencioni I've updated the code to log in the default reporter, though I'm still logging each individual linting as a period in the runner. I'm not familiar at all with the ecosystem, however. Does this work?

lib/scss_lint/runner.rb
```diff
@@ -15,9 +17,10 @@ def initialize(config)
     # @param files [Array<Hash>] list of file object/path hashes
     def run(files)
       @files = files
-      @files.each do |file|
+      result = @files.each do |file|
```
sds: What are you trying to do here? `result` will be set to `@files` since you're using `each`...

sds:
Hey @adsteel,

No, this does not work. If I run `scss-lint --format JSON`, I get output like the following:

```
..................................................{
  "test.scss": [
    {
      "line": 1,
      "column": 2,
      "length": 140,
      "severity": "warning",
      "reason": "Variable names should be followed immediately by a colon",
      "linter": "SpaceAfterVariableName" 
    }
  ]
}
```

...which will obviously break any tools that rely on this format ([Overcommit](https://github.com/brigade/overcommit) being one of them).

This needs to support writing output to a separate stream (STDERR) so that the output format isn't broken.

adsteel:
@sds Okay, I'll take one more swing at this on Friday. Without more knowledge of the ecosystem, or tests that define the API spec, this is a little like groping around in the dark.

adsteel:
I wasn't able to come up with a solution I liked. I may re-attempt this at some point if someone else doesn't solve it, but in the meantime I need to stop working on it.

PR #710, opened 2016-02-12
Visit children in SingleLinePerProperty

### LINKED ISSUE #709: SingleLinePerProperty gets confused with nesting
With the following configuration:

```yaml
SingleLinePerProperty:
  enabled: true
  allow_single_line_rule_sets: false
```

And the following SCSS:

```scss
.test { padding: 0; margin: 0; }
```

The SingleLinePerProperty correctly reports a lint here. However, if this is nested:

```scss
.testing {
  .test { padding: 0; margin: 0; }
}
```

it is quiet.

lencioni:
With the following configuration:

```yaml
SingleLinePerProperty:
  enabled: true
  allow_single_line_rule_sets: false
```

And the following SCSS:

```scss
.test { padding: 0; margin: 0; }
```

The SingleLinePerProperty correctly reports a lint here. However, if
this is nested:

```scss
.testing {
  .test { padding: 0; margin: 0; }
}
```

it is quiet.

It seems that the Sass parser is not visiting the child rule here. I
don't know enough about how this is supposed to work to know if that is
expected or not, but it seems that I can make this work by adding a line
that will recurse the children.

Fixes #709

### DIFF
--- a/lib/scss_lint/linter/single_line_per_property.rb
+++ b/lib/scss_lint/linter/single_line_per_property.rb
@@ -4,6 +4,8 @@ module SCSSLint
     include LinterRegistry
 
     def visit_rule(node)
+      yield # Continue linting children
+
       single_line = single_line_rule_set?(node)
       return if single_line && config['allow_single_line_rule_sets']
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
   list/map literals
 * Improve compatibility with files that use CRLF and CR for newlines.
 * Load `plugin_directories` relative to the config file itself
+* Fix bug in `SingleLinePerProperty` that prevented the linter from checking
+  nested selectors
 
 ## 0.44.0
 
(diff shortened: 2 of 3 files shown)


lib/scss_lint/linter/single_line_per_property.rb
```diff
@@ -4,6 +4,8 @@ class Linter::SingleLinePerProperty < Linter
     include LinterRegistry
 
     def visit_rule(node)
+      visit_children(node)
```
sds: This can just be `yield`, to match what we do elsewhere. It's perhaps not as descriptive but at least consistent. I usually write:

```ruby
yield # Continue linting children
```

sds:
:+1:

lib/scss_lint/linter/single_line_per_property.rb
lencioni: Ah, I guess that's why my grepping for `visit_children` didn't really turn up much.

PR #713, opened 2016-02-16
Relax rake dependency version to allow >= 0.9

lencioni:
I am working in a project that is running an old version of rake, and
for historical reasons, it is going to be very difficult to upgrade to
rake 10. I tried looking into why this was set at rake 10 and I didn't
really come up with much, so I am relaxing this version requirement a
little.

I tried this out locally and specs pass on 0.9.* but not on 0.8.*, so
that seems like a good place to draw the line.

### DIFF
--- a/scss_lint.gemspec
+++ b/scss_lint.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = '>= 1.9.3'
 
-  s.add_dependency 'rake', '~> 10.0'
+  s.add_dependency 'rake', '>= 0.9', '< 11'
   s.add_dependency 'sass', '~> 3.4.15'
 
   s.add_development_dependency 'rspec', '~> 3.4.0'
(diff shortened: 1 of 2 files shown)


scss_lint.gemspec
```diff
@@ -25,7 +25,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = '>= 1.9.3'
 
-  s.add_dependency 'rake', '~> 10.0'
+  s.add_dependency 'rake', '>= 0.9', '~> 10.0'
```
sds: Should this be:

```ruby
s.add_dependency 'rake', '>= 0.9', '< 11'
```

My understanding is that all constraints must be satisfied, so `~> 10.0` will sill require you to have rake >= 10.0.

lencioni: Ah, that could be. I'll update it.

lencioni:
Updated!

sds:
:+1:

Issue #708, opened 2016-02-11
BorderZero message should match line


lencioni:
On lines that do not follow the BorderZero linter configuration but use a property like `border-top`, we currently have a message that says something like "`border: 0` is preferred over `border: none`". It would be much nicer if this used the actual property being used here instead of just `border` every time.

cih:
That sounds like a nice improvement and should be fairly straightforward, I'll have a go at this and push up a PR shortly.

PR #714, opened 2016-02-16
Improve BorderZero lint description

cih:
This change updates the `BorderZero` linter
by adding the offending border property to
the lint description.

Fixes https://github.com/brigade/scss-lint/issues/708


lencioni:
Only one small comment from me, otherwise this looks great! Thanks for the PR.

Issue #114, opened 2014-04-04
DuplicateProperty is linting fallbacks


medoingthings:
I'm having an issue where SCSSLint warns me like this:

```
Running "scsslint:all" (scsslint) task
Running scss-lint on all
hero.scss:139 [W] Property `background-color` already defined on line 138
```

SCSS:

```
.classname {
    background-color: #fff;
    background-color: rgba(255, 255, 255, 0.7);
}
```

According to the [documentation](https://github.com/causes/scss-lint/blob/master/lib/scss_lint/linter/README.md#duplicateproperty) it should pass.

Anybody else having the same issue?

polybuildr:
CSSLint [does this in an interesting way](https://github.com/CSSLint/csslint/wiki/Disallow-duplicate-properties) by ignoring duplicate properties if they're on consecutive lines.

PCateNumbersUSA:
I like the way CSSLint does this as @polybuildr mentioned. Added this as an option would be great.

PR #717, opened 2016-02-17
Configure to not lint DuplicateProperty on consecutive lines

### LINKED ISSUE #715: DuplicateProperty consecutive duplicates
In #114, after it was closed, a discussion began regarding how CSSLint [ignores consecutive duplicates](https://github.com/CSSLint/csslint/wiki/Disallow-duplicate-properties). The idea is that if you write:

```css
a {
  background: #fff;
  background: rba(255, 255, 255, .5);
}
```

then you probably did so intentionally, since its visually so easy to see that the same property is defined twice. If you hadn't meant this to happen, you would have seen it and fixed it.

I'd be happy to contribute a config for DuplicateProperty, like `ignore_consecutive` or `allow_consecutive`, or something.

srawlins:
Fixes #715. CC #114.

Explanation added to README.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # SCSS-Lint Changelog
 
-## master (unreleased)
+## 0.45.0
 
 * Add `outline-{color,style,width}` properties to `recess` sort order list
 * Fix `TrailingSemicolon` to not report false positives for single line
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # SCSS-Lint Changelog
 
+## master (unreleased)
+
+* Add `ignore_consecutive` option to the `DuplicateProperty` linter, allowing
+  duplicate consecutive properties. It accepts `true`, `false`, or a list of
+  property names to be ignored.
+
 ## 0.45.0
 
 * Add `outline-{color,style,width}` properties to `recess` sort order list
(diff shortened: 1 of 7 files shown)


lencioni:
Do you think it would be a good idea to make the ignore option be a whitelist instead of a boolean? Or at least use a whitelist internally? I ask because if you are one to sort your properties, you might end up with consecutive duplicates when you don't actually want them but won't be warned. Restricting this to a whitelist seems like a good way to go, to minimize this problem.

srawlins:
I'd be happy to change it to that. Or maybe we could have both features.

I like the consecutive lines solution because I think it has _less_ chance of accidentally allowing a duplicate than the whitelist solution. The whitelist solution can definitely accidentally allow a duplicate of `background` where the duplicate is unintentional.

The consecutive solution, however, only allows duplicates by accident if the humans also missed a duplicate consecutive property.

lencioni:
Oh, I meant using a whitelist in addition to the consecutive check.

srawlins:
Oh I see. So you would specify something like:

```
  ignore_consecutive:
    - background
    - transition
```

I like this idea.

lencioni:
Yeah, something like that. I think perhaps it should come with a default list of properties built in that commonly are used with fallbacks, so it might need to use two options--one to turn on/off this behavior and another to customize the list.

lencioni:
However, if you are using something like `calc` and want fallbacks for that, you might need to include a lot of properties in your whitelist. Perhaps the whitelist should look at values instead of properties? But that gets hard if you are using a Sass function...

srawlins:
Awesome, pushed and squashed. I didn't look into the *values* whitelist. Too complicated :grinning:

lencioni:
Besides the comment about the code block, this looks good to me. Want to fix that and add a changelog entry?

srawlins:
Thanks! Done

Issue #646, opened 2015-12-01
Feature Request: disallow global variables


janpaul123:
SCSS is not great with local scoping, but there could be a combination of rules that would prevent accidentally creating dependencies between SCSS files. For example:

1. A rule that says that variables prefixed with an underscore must be defined in the same file (to enforce that they are local variables). E.g. `$_background-color`
2. A rule that disallows creating variables that are not prefixed with an underscore, except when you explicitly disable that rule in certain files that contain a bunch of global variables.

What do you think?

sds:
Sounds reasonable to me!

PR #722, opened 2016-02-18
Add PrivateNamingConvention linter

### LINKED ISSUE #721: Enforce privacy for leading underscore naming convention
Although the `NameFormat` linter has an option to allow for leading underscores, it does not enforce the convention. It would be really great to either have a linter that does this, or teach `NameFormat` to do this.

What I mean by enforcing privacy for the leading underscore convention is, for variables, functions, and mixins that start with a leading underscore, we should verify that:

- any definitions are used within the same file
- any uses are defined within the same file

And potentially:

- any uses are preceded by their definitions

**Bad**
```scss
$_foo: red;

p {
  color: blue;
}
```

**Bad**
```scss
p {
  color: $_foo;
}
```

**Bad**
```scss
p {
  color: $_foo;
}

$_foo: red;
```

**Good**
```scss
$_foo: red;

p {
  color: $_foo;
}
```

And so on for functions and mixins.

lencioni:
Although the `NameFormat` linter has an option to allow for leading
underscores, it does not enforce the convention. It would be really
great to either have a linter that does this.

What I mean by enforcing privacy for the leading underscore convention
is, for variables, functions, and mixins that start with a leading
underscore, we should verify that:

- any definitions are used within the same file
- any uses are defined within the same file
- any uses are preceded by their definitions

I considered adding this to `NameFormat`, but it seems specialized
enough that I decided to put it in its own linter. I left the prefix
configurable but defaulting to `_`, which is likely what most people
would use anyway.

Fixes #721

### DIFF
--- a/lib/scss_lint/linter/private_naming_convention.rb
+++ b/lib/scss_lint/linter/private_naming_convention.rb
@@ -1,0 +1,86 @@
+module SCSSLint
+  # Verifies that variables, functions, and mixins that follow the private naming convention are
+  # defined and used within the same file.
+  class Linter::PrivateNamingConvention < Linter
+    include LinterRegistry
+
+    def visit_root(node)
+      # Register all top-level function, mixin, and variable definitions.
+      node.children.each_with_object([]) do |child_node|
+        if child_node.is_a?(Sass::Tree::FunctionNode)
+          register_node child_node, 'function'
+        elsif child_node.is_a?(Sass::Tree::MixinDefNode)
+          register_node child_node, 'mixin'
+        elsif child_node.is_a?(Sass::Tree::VariableNode)
+          register_node child_node, 'variable'
+        else
+          yield
+        end
+      end
+
+      # After we have visited everything, we want to see if any private things
+      # were defined but not used.
+      after_visit_all
+    end
+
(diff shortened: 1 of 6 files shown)


lib/scss_lint/linter/private_naming_convention.rb
```diff
@@ -0,0 +1,86 @@
...
+      node.children.each_with_object([]) do |child_node|
+        if child_node.is_a?(Sass::Tree::FunctionNode)
+          register_node child_node, 'function'
+        elsif child_node.is_a?(Sass::Tree::MixinDefNode)
+          register_node child_node, 'mixin'
+        elsif child_node.is_a?(Sass::Tree::VariableNode)
+          register_node child_node, 'variable'
+        else
+          yield
+        end
```
sds: This would read better as a `case` statement:

```ruby
case child_node
when Sass::Tree::FunctionNode
  ...
  ...
else
  ...
end
```

sds:
Nice! This also seems to address #646. :+1:

lib/scss_lint/linter/private_naming_convention.rb
lencioni: Done!

PR #723, opened 2016-02-19
Configure to not lint DuplicateProperty on consecutive lines

### LINKED ISSUE #715: DuplicateProperty consecutive duplicates

srawlins:
Fixes #715. CC #114.

This includes a whitelist concept. The new `ignore_consecutive` option can be
specified with a list.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # SCSS-Lint Changelog
 
+## master (unreleased)
+
+* Add `ignore_consecutive` option to the `DuplicateProperty` linter, allowing
+  duplicate consecutive properties. It accepts `true`, `false`, or a list of
+  property names to be ignored.
+
 ## 0.45.0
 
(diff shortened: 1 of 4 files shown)


srawlins:
Sorry, #717 got borked when I was trying to squash commits. This should be ready to go.

lencioni:
:+1:

PR #724, opened 2016-02-20
Fix locally failing LinterGem spec

lencioni:
This spec was failing locally for me because File.exist? was being
called with unexpected arguments, and we had stubbed it but not allowed
these unexpected arguments. I fixed this by just having it call through
to the original in this case.

### DIFF
--- a/spec/scss_lint/plugins/linter_gem_spec.rb
+++ b/spec/scss_lint/plugins/linter_gem_spec.rb
@@ -10,6 +10,7 @@ module SCSSLint
       let(:config_file_exists) { false }
 
       before do
+        File.stub(:exist?).and_call_original
         File.stub(:exist?).with(config_file).and_return(config_file_exists)
       end
 


sds:
Interesting. What version of Ruby are you running locally? I can't reproduce the failure myself (and Travis isn't complaining), so I'm really curious why this is necessary for you locally. 

I'm not strongly against merging this, but non-reproducible failures are usually indicative of some other problem.

lencioni:
Yeah, I thought it was weird.

ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-darwin14]

lencioni:
Here's the failure:

```
1) SCSSLint::Plugins::LinterGem#load when the gem does not exist raises an exception
   Failure/Error: expect { subject.load }.to raise_error Exceptions::PluginGemLoadError

     expected SCSSLint::Exceptions::PluginGemLoadError, got #<RSpec::Mocks::MockExpectationError: #<File (class)> received :exist? with unexpected arguments
       expected: ("/gem_dir/.scss-lint.yml")
            got: ("/Users/joe/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/extensions/x86_64-darwin-14/2.3.0-static/ffi-1.9.10/gem.build_complete")
      Please stub a default value first if message might be received with other args as well.
     > with backtrace:
       # ./lib/scss_lint/plugins/linter_gem.rb:12:in `load'
       # ./spec/scss_lint/plugins/linter_gem_spec.rb:18:in `block (5 levels) in <module:SCSSLint>'
       # ./spec/scss_lint/plugins/linter_gem_spec.rb:18:in `block (4 levels) in <module:SCSSLint>'
   # ./spec/scss_lint/plugins/linter_gem_spec.rb:18:in `block (4 levels) in <module:SCSSLint>'
```

lencioni:
So it looks like the `ffi` gem is doing something here.

lencioni:
Which looks like it might be a dependency of a dependency of a dependency of the travis gem. Do you get this if you install the travis gem?

sds:
I'm still not able to reproduce using Ruby 2.3.0 on a Mac, even after installing `travis` (which is already installed from the `Gemfile`, but I tried installing the latest version anyway):

```
⨠ ruby -v
ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-darwin14]
⨠ gem install travis
Fetching: travis-1.8.2.gem (100%)
Successfully installed travis-1.8.2
1 gem installed
⨠ bundle exec rspec
..............................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................

Finished in 4.65 seconds (files took 0.77626 seconds to load)
1566 examples, 0 failures
```

lencioni:
I think it might be because I was running `rspec` without `bundle exec`. Are you able to reproduce that way?

FWIW, here's what I get on a different machine:

```
1) SCSSLint::Plugins::LinterGem#load when the gem does not exist raises an exception
   Failure/Error: expect { subject.load }.to raise_error Exceptions::PluginGemLoadError

     expected SCSSLint::Exceptions::PluginGemLoadError, got #<RSpec::Mocks::MockExpectationError: #<File (class)> received :exist? with unexpected arguments
       expected: ("/gem_dir/.scss-lint.yml")
            got: ("/Users/joe_lencioni/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/extensions/x86_64-darwin-14/2.3.0-static/bcrypt-3.1.9/gem.build_complete")
      Please stub a default value first if message might be received with other args as well.
     > with backtrace:
       # ./lib/scss_lint/plugins/linter_gem.rb:12:in `load'
       # ./spec/scss_lint/plugins/linter_gem_spec.rb:18:in `block (5 levels) in <module:SCSSLint>'
       # ./spec/scss_lint/plugins/linter_gem_spec.rb:18:in `block (4 levels) in <module:SCSSLint>'
   # ./spec/scss_lint/plugins/linter_gem_spec.rb:18:in `block (4 levels) in <module:SCSSLint>'
```

ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-darwin14]

sds:
I can indeed reproduce once I run without `bundle exec`.

I'm not sure we want to support running the tests outside a Bundler context. When you do that you throw away any guarantee of which dependencies you load, so you might get spurious failures which you spend way too much time debugging that would not actually affect anyone.

If you'd like to support running `rspec` without `bundle exec`, you can instantiate the Bundler context manually at the top of the `spec/spec_helper.rb` file:

```ruby
# spec/spec_helper.rb
require 'bundler/setup'
```

This should ensure appropriate gem versions are in your `$LOAD_PATH`. I tend to avoid this because it is "magic", but I'm open to it here since it's harmless if you still use `bundle exec ...` to run it.

lencioni:
Yeah, I think it is fine to not support running without `bundle exec`. Closing this out.

PR #727, opened 2016-02-22
Add TAP reporter

### LINKED ISSUE #698: Add TAP reporter
To provide better compatibility with more tools, it would be nice to add a TAP reporter to SCSS-Lint.

http://testanything.org/

lencioni:
To provide better compatibility with more tools, I thought that it would
be nice to add a TAP reporter to SCSS-Lint.

  http://testanything.org/

For a list of TAP consumers, look at

  http://testanything.org/consumers.html

Fixes #698


sds:
Nice, thanks for the sweet tests! :+1:

PR #728, opened 2016-02-22
Improve nested private names in PrivateNamingConvention

lencioni:
I was thinking about this rule and realized an edge case that I had
forgotten to account for. What happens if you have a nested private
variable? This wasn't handled very well, but this commit improves it.

### DIFF
--- a/lib/scss_lint/linter/private_naming_convention.rb
+++ b/lib/scss_lint/linter/private_naming_convention.rb
@@ -7,11 +7,12 @@ module SCSSLint
     def visit_root(node)
       # Register all top-level function, mixin, and variable definitions.
       node.children.each_with_object([]) do |child_node|
-        if child_node.is_a?(Sass::Tree::FunctionNode)
+        case child_node
+        when Sass::Tree::FunctionNode
           register_node child_node, 'function'
-        elsif child_node.is_a?(Sass::Tree::MixinDefNode)
+        when Sass::Tree::MixinDefNode
           register_node child_node, 'mixin'
-        elsif child_node.is_a?(Sass::Tree::VariableNode)
+        when Sass::Tree::VariableNode
           register_node child_node, 'variable'
         else
           yield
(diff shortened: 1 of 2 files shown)


lib/scss_lint/linter/private_naming_convention.rb
```diff
@@ -82,5 +131,27 @@ def after_visit_all
...
+    def defining_class_for(node)
+      case node
+      when Sass::Script::Tree::Funcall
+        Sass::Tree::FunctionNode
+      when Sass::Tree::MixinNode
+        Sass::Tree::MixinDefNode
+      when Sass::Script::Tree::Variable
+        Sass::Tree::VariableNode
+      end
+    end
```
sds: You can combine these into a single hash, i.e. repurpose `DEFINING_CLASSES` into something like (forgive the naming):

```ruby
DEFINITIONS = {
  Sass::Tree::FunctionNode => { use_type: Sass::Script::Tree::Funcall, human_name: 'function' },
  ...,
}.freeze

DEFINITION_NAMES = DEFINITIONS.map { |k v| [k, v[:human_name] }.freeze
DEFINITION_TYPES = DEFINITION_TYPES.map { |k, v| [v[:use_type], k] }.freeze
```

Then these implementations become:

```ruby
def humanize_node_class(node)
  DEFINITION_NAMES[node]
end

def defining_class_for(node)
  DEFINITION_TYPES[node]
end
```

The advantage being you define all types once, rather than having to list these out in multiple places in verbose `case` statements. Not a big deal.

sds:
:+1:

lib/scss_lint/linter/private_naming_convention.rb
lencioni: Sounds like a plan

lencioni: https://github.com/brigade/scss-lint/commit/8106117625a761962d3c5caba56644d442b419ae

PR #735, opened 2016-02-23
Fix TAP reporter format

lencioni:
When I implemented this, I closed the YAML block with --- when it is
supposed to be closed with ...


sds:
Merged in 9d4598f.

PR #736, opened 2016-02-23
Add a preprocess_command option.

### LINKED ISSUE #689: What to do about Jekyll assets, with triple dashes
Jekyll [specifies](http://jekyllrb.com/docs/assets/) that asset files, such as a .scss file, should start with two lines of triple dashes:

```
---
---
```

for some reason or other. Of course, these two lines make this an invalid SCSS file, which means the Sass gem can't parse it, which means we can't lint it. Anything to be done?

srawlins:
Fixes #689 

Annoying that this fix is for such a trivial issue. But possibly preprocessors give lots of power to users? See the README changes for a description of the new `preprocess_command` and `preprocess_files` options.

Here's a list of design decisions that you may particularly want to review:

* I added a top-level section to the README, because it's pretty big.
* I made a new PreprocessorError, which will cause scss-lint to exit with a new code, 84.
* I essentially doubled the number of options that `SCSSLint::Engine` takes, which I feel bad about. Like a slippery slope kind of thing. Let me know if that interface should be refactored.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ it into your [SCM hooks](https://github.com/brigade/overcommit).
 * [Exit Status Codes](#exit-status-codes)
 * [Linters](#linters)
 * [Custom Linters](#custom-linters)
+* [Preprocessing](#preprocessing)
 * [Editor Integration](#editor-integration)
 * [Git Integration](#git-integration)
 * [Rake Integration](#rake-integration)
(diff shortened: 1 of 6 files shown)


sds:
Thanks for writing thorough documentation and tests, @srawlins!

Merged in f3d0f947fca27ffebffdd9aca797d16cafe96693. Happy to have a generalized solution for this problem. I'm not too worried about the options for the time being, though I appreciate your pointing it out.

PR #744, opened 2016-03-02
Fix bug with TAP reporter using file hashes as filenames

lencioni:
When I wrote this reporter, I assumed that the `files` was an array of
strings (partly because that is how it was documented). However, in
practice, it is an array of hashes, so this didn't work correctly.


lencioni:
@sds I have tested this locally and it is working correctly now. Any chance we can get a patch release out today?

lencioni:
I have another small commit in the works on top of this one, FYI

PR #761, opened 2016-03-18
Add linear-gradient() to NameFormat whitelist.

abea:
This change adds the CSS function `linear-gradient()` to the NameFormat whitelist. Since it's not a Sass function it shouldn't be throwing a lint warning.


sds:
Thanks!

Merged in 5cb125bc60cba34c8cad409f6061a0e869a50799

PR #762, opened 2016-03-19
Change DisableLinterReason logic

### LINKED ISSUE #749: DisableLinterReason not catching a comment for an inline disable
I have the following SCSS:

```scss
// Unfortunately this selector is necessary for CSS specificity reasons.
> .notifications-container ul > li:last-of-type { // scss-lint:disable SelectorDepth
     border-bottom: 0;
}
```

I remember fixing the SCSS linter reporting this unfortunately necessary selector with the above. Adding it as a comment at the end of a line was documented somewhere, but I can't find it now. It's supposed to disable the rule for only that line.

The `DisableLinterReason` rule is throwing a warning despite the fact that I have a comment explanation up above.

srawlins:
Fixes #749 

I tried to keep the code as readable as possible, with as few double negatives as possible :smirk:

### DIFF
--- a/lib/scss_lint/linter/disable_linter_reason.rb
+++ b/lib/scss_lint/linter/disable_linter_reason.rb
@@ -7,19 +7,37 @@ module SCSSLint
       # No lint if the first line of the comment is not a command (because then
       # either this comment has no commands, or the first line serves as a the
       # reason for a command on a later line).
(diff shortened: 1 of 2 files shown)


sds:
Thanks for the fix!

Merged in 9fb936a.

PR #777, opened 2016-04-15
Add GitLab to INTHEWILD.md

connorshea:
We use SCSS Lint for GitLab: https://gitlab.com/gitlab-org/gitlab-ce/blob/master/.scss-lint.yml


lencioni:
Thanks!

PR #783, opened 2016-04-26
Added known Sass and CSS functions to whitelist.

abea:
Upon running up against another core function getting NameFormat warning, I added all documented [Sass functions](http://sass-lang.com/documentation/Sass/Script/Functions.html) that could trigger the warning. Also added three additional CSS functions I'm aware of: `repeating-linear-gradient`, `radial-gradient`, `repeating-radial-gradient`.

### DIFF
--- a/lib/scss_lint/linter/name_format.rb
+++ b/lib/scss_lint/linter/name_format.rb
@@ -39,7 +39,19 @@ module SCSSLint
       scaleX scaleY scaleZ
       skewX skewY
       translateX translateY translateZ
-      linear-gradient
+      linear-gradient repeating-linear-gradient
+      radial-gradient repeating-radial-gradient
+      adjust-hue adjust-color scale-color change-color ie-hex-str
+      str-length str-insert str-index str-slice to-upper-case to-lower-case
+      list-separator
+      map-get map-merge map-remove map-keys map-values map-has-key
+      selector-nest selector-append selector-extend selector-replace
+      selector-unify is-superselector simple-selectors selector-parse
+      feature-exists variable-exists global-variable-exists function-exists
+      mixin-exists type-of
+      unique-id
+
+
     ].to_set
 
     def check_name(node, node_type, node_text = node.name)


lib/scss_lint/linter/name_format.rb
```diff
@@ -39,7 +39,19 @@ def visit_variable(node)
...
+      str-length str-insert str-index str-slice to-upper-case to-lower-case
+      list-separator
+      map-get map-merge map-remove map-keys map-values map-has-key
+      selector-nest selector-append selector-extend selector-replace
+      selector-unify is-superselector simple-selectors selector-parse
+      feature-exists variable-exists global-variable-exists function-exists
+      mixin-exists type-of
+      unique-id
+
+
```
lencioni: Looks like CI failed because of these two blank lines. Will you remove them?

lib/scss_lint/linter/name_format.rb
```diff
@@ -39,7 +39,19 @@ def visit_variable(node)
       scaleX scaleY scaleZ
       skewX skewY
       translateX translateY translateZ
-      linear-gradient
+      linear-gradient repeating-linear-gradient
```
lencioni: Since you are adding sass specific functions and all we previously had was CSS functions, it might be nice to use a separate set for these to keep things clear.

sds:
Thanks for the pull request, @abea.

It might make more sense to query the list of Sass functions directly, e.g.

```ruby
> puts Sass::Script::Functions.instance_variable_get(:@signatures).keys.grep(/_/).map { |name| name.to_s.tr('_', '-') }.sort
adjust-color
adjust-hue
...
unique-id
variable-exists
```

This is a form of tight coupling, but this project is already pretty tightly-coupled with the Sass gem anyway, given the parse tree hacks it relies on.

abea:
Thanks for the comments, guys. I'll incorporate those and update this as soon as I can.

sds:
Closing due to lack of activity, but would be happy to merge support for these additional functions. 

Feel free to submit a new pull request when it's ready!

PR #785, opened 2016-04-26
Feature/stdout total numbers of errors

mir4a:
It would be nice to have a total number of errors/warnings after each run:
screen shot 2016-04-26 at 22 24 13


sds:
As mentioned in #695, modifying the output in this way breaks other formatters. Any attempt at changing the default output needs to not affect the output of other formatters.

Feel free to open another pull request that addresses this concern. Thanks!

PR #788, opened 2016-04-28
Add css properties to smacss.txt

danny-andrews:
Add  `clip` and `background-clip`


sds:
Merged in a1cd658. Thanks!

PR #800, opened 2016-05-25
Add Length Variable linting similar to Color Variable

dnl:
As an experiment I'm requiring all lengths (numbers with length units) to be variables, because often lengths are related to other lengths and variables prevent them being magical magic numbers.

There are a couple of limitations. 
- I haven't flagged percentage values, a) because that already expresses how lengths are related to each other, and b) because percentages are also used for non-length values. (e.g. `fade(black, 70%)`)
- I'd prefer `width: $button-width + 2;` was flagged, because I want to know where the '2' comes from. sass is too clever here.

I've also made it so you can configure specific lengths (e.g. `1px`) to never be flagged, and specific properties (e.g. `text-shadow`)
And while I was at it I made a similar configuration for colors.

Is this something you'd be interested in merging? If so I'll add to documentation, etc.

### DIFF
--- a/lib/scss_lint/linter/color_variable.rb
+++ b/lib/scss_lint/linter/color_variable.rb
@@ -36,9 +36,15 @@ module SCSSLint
   private
 
     def record_lint(node, color)
+      return if permitted_color?(color)
+
       add_lint node, "Color literals like `#{color}` should only be used in " \
                      'variable declarations; they should be referred to via ' \
                      'variable everywhere else.'
+    end
+
+    def permitted_color?(color)
+      return true if config['allowed_colors'] && config['allowed_colors'].include?(color)
     end
 
     def literal_string?(script_string)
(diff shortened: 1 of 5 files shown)


lib/scss_lint/linter/color_variable.rb
```diff
@@ -36,11 +36,17 @@ def visit_script_funcall(node)
...
+
       add_lint node, "Color literals like `#{color}` should only be used in " \
                      'variable declarations; they should be referred to via ' \
                      'variable everywhere else.'
     end
 
+    def permitted_color?(color)
+      return true if config['allowed_colors'] && config['allowed_colors'].include?(color)
+    end
+
```
sds: The changes to this file seem unrelated. Can you remove them?

sds:
Thanks for the pull request @dnl!

Since new linters are disabled by default, I'm fine including this so long as the relevant documentation is included. Users who want to opt-in to this can easily do so should they chose.

Small comment about a change that seems unrelated, but otherwise looks good!

sds:
Thanks for adding documentation, @dnl!

Merged in a1a6faa8e762df22dda04601cad4378d0c07d57c.

PR #801, opened 2016-05-25
Nameformat whitelist

abea:
Adds known SCSS and CSS functions to the NameFormat whitelist, separating those into two lists. Since it now checks two lists, the check has been moved into the `whitelist?` method.

At the recommendation of a Ruby developer colleague (I'm not one) I've also frozen the whitelists to prevent run-time changes as well as moved the whitelist variables out of the private section since it seems that has no effect on those array variables and could be misleading.


sds:
Thanks for the pull request, @abea!

Merged in 1f1381142d56afd406aa27ad4e5bef44d5e3ec77.

PR #807, opened 2016-07-01
Wrap at-declaratives in backticks

teoljungberg:
We, at thoughtbot, use `scss-lint` to check for style violations in
[Hound]. Hound will comment on style violations on an open PR,
and leave comments on the violating lines.

To avoid pinging GitHub users in the comment we can wrap the
declaratives in backticks.

This was done prior in https://github.com/brigade/scss-lint/pull/354

[Hound]: https://houndci.com/


sds:
Thanks!

else:
:+1: thanks a lot!

PR #819, opened 2016-08-02
Fix linter_options to actually allow setting options

### LINKED ISSUE #718: cli parameter "--exclude-linter" not working on version 0.44.0
The title says it all.  
I get no errors, it just doesn't exclude any linters that I place in the parameter. The yml file works, but the cli command does not.  I think this is the same for "--include-linter".
However, reverting back to version 0.38.0 works fine.

`gem uninstall scss_lint`
`gem install scss-lint -v 0.38.0`

I am using Mac OSX 10.11

ibrahima:
The way it's currently phrased, each time you call linter_options it
returns a new hash, because it's merging a hash literal with the
existing options. Thus you can never change the options.

Note: This doesn't take care of setting the value of
`@options['linters'][linter_name]` to the new hash if a new hash was
returned by fetch, but neither does the existing code so I'm not sure if
that's also desired.

This fixes #718 and potentially other issues with setting linter specific options.

### DIFF
--- a/lib/scss_lint/config.rb
+++ b/lib/scss_lint/config.rb
@@ -270,9 +270,9 @@ module SCSSLint
     end
 
     def linter_options(linter)
-      { 'severity' => @options['severity'] }.merge(
-        @options['linters'].fetch(self.class.linter_name(linter), {})
-      )
+      options = @options['linters'].fetch(self.class.linter_name(linter), {})
+      options['severity'] ||= @options['severity']
+      options
     end
 
     def excluded_file?(file_path)


connorshea:
@ibrahima thank you! 👏

sds:
Thanks for digging into this issue, @ibrahima.

Looking at the [commit](https://github.com/brigade/scss-lint/commit/f6061d1d18210216e6ec438372d871f6efc0ed99) that previously changed this, I can't see/recall why that particular approach was taken. Hopefully this doesn't have other unintended consequences.

Thanks again!

ibrahima:
No problem, thank you!

The hash literal+merge approach seems to be a common idiom for default parameters in ruby, but in this case you're not just generating default parameters, but also getting existing ones and potentially changing them, so it doesn't quite work the way you'd want.

Issue #813, opened 2016-07-18
Linter Proposal: LeadingCommentSpace


ivantsepp:
I would like to propose a linter that's similar to what [Rubocop](http://www.rubydoc.info/github/bbatsov/rubocop/Rubocop/Cop/Style/LeadingCommentSpace) has. It would throw a warning if there is no leading space in comments.

```scss
//bad comment
/*bad comment */
```

```scss
// good comment
/* good comment */
```

Let me know what you think! I would like to also implement this linter if you agree with this =)

sds:
Implemented in #825

PR #825, opened 2016-08-16
Add SpaceAfterComment Linter

ivantsepp:
Adds a linter for #813.

This is a big PR (alot of it is tests) so definitely let me know what changes you would like to see in order to get this merged! (or if it should be merged at all). 

The basic idea is to look at comment nodes and depending on the type, we do different things. 
 - If it's `:silent`, we need to split by new lines since sass will combine consecutive comments `//` into one node. There is also an option to allow empty `//` comments.
 - If it's `:normal`, count the spaces after `/*`
 - If it's `:loud`, count the spaces after `/*!`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Fix `--exclude-linter` flag
 * Modify `ColorKeyword` linter to allow color keywords to be used as arguments
   in `map-*`-related function calls
+* Add `SpaceAfterComment` which checks for spacing after comment literal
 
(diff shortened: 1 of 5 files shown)


sds:
Thanks for the detailed pull request, @ivantsepp. 

Including thorough tests and documentation makes this much easier to 👍 

Merged in 21d244440408af65a1c696ea7c99b45218fa0b7f.

ivantsepp:
Thanks for reviewing and merging!

PR #826, opened 2016-08-17
SpaceAfterComma linter fixes

ivantsepp:
For this example:

```scss
.test {
  color: rgba(0, 0,0,.1);
}
```

with the following configuration:

```
  SpaceAfterComma:
    enabled: true
    style: no_space
```

You get the following linter messages:

```
test.scss:2:1 [W] SpaceAfterComma: Commas in function arguments should be followed by a single space
test.scss:2:1 [W] SpaceAfterComma: Commas in function arguments should be followed by a single space
```

The two things I noticed here is:
 - the column number in the location is `1` (meaning the column number is set)
 - the linter message is set to `followed by a single space` regardless of your setting. 

This PR addresses both of these issues. With the changes, this is the output you would now get:

```
test.scss:2:19 [W] SpaceAfterComma: Commas in function arguments should be followed by one space
test.scss:2:21 [W] SpaceAfterComma: Commas in function arguments should be followed by one space
```

### DIFF
--- a/lib/scss_lint/linter.rb
+++ b/lib/scss_lint/linter.rb
@@ -52,7 +52,8 @@ module SCSSLint
 
     # Helper for creating lint from a parse tree node
     #
-    # @param node_or_line_or_location [Sass::Script::Tree::Node, Fixnum, SCSSLint::Location]
+    # @param node_or_line_or_location [Sass::Script::Tree::Node, Fixnum,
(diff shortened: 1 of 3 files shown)


sds:
Thanks @ivantsepp!

Merged in a229ca164ba028fd2ecc5f6ce256deed706ffd35.

PR #853, opened 2016-10-14
Linking to Bamboo-visualization addon at checkstyle-formatter

SchulteMarkus:


sds:
Hey @SchulteMarkus, thanks for the pull request.

Unless I'm missing something, Bamboo is a proprietary solution with no free offering for open source projects. It has a 30 day free trial but that's not the same as a truly free open source offering like Jenkins.

With this in mind, I'm uncomfortable adding what is effectively free advertising to SCSS-Lint's documentation for a proprietary product. Hope that's understandable. Again, if there's something here I'm missing, do let me know.

Thanks!

SchulteMarkus:
Hey @sds,

you are right about Bamboo - it's not open source, neither is it free at all. So I do understand your opinion, but I am having a different one. Of course, it is your good right not to accept this PullRequest, but maybe I can change your mind.
In my opinion Bamboo or the addon (I am author of!) itself should be listed here, because this platform/Wiki/ReadME is about information. Each reader can and will decide on it's own, what to do with this information, but without the information, the ReadME is a litte bit less useful. As an example: Imagine, Wikipedia would not list Microsoft. Or my thesis as a question: Wouldn't a Bamboo user benefit from this link to the Checkstyle-addon in the Atlassian marketpalce?

chris-dura:
Following that rationale would cause every tool that uses a Checkstyle plugin/addon to be listed in this project's README, which obviously is not practical. The Checkstyle scss-lint formatter is germane to the scss-lint tool itself; whereas, every tool that has a Checkstyle plugin/implementation are not. I think @sds intent was just to provide an example of how a user might use the output, not to create an exhaustive list of workflows and tools that could use it.

I think simply _knowing_ that `scss-lint` can output Checkstyle format is enough for users who want to use Checkstyle in their tooling.

Issue #787, opened 2016-04-27
Unable to lint files with YAML front matter


shawnbot:
I have a `.css` file in a Jekyll site that's structured like this:

```scss
---
# this needs YAML front matter to be compile by Jekyll's Sass
---
@import 'variables';

.blah {
  color: $red;
}
```

[Jekyll](https://jekyllrb.com/docs/assets/) dutifully strips the front matter and passes the rest off to Sass, but both the `sass` CLI and `scss-lint` barf on it:

```
css/blah.scss:5: Invalid CSS after "...ort 'variables'": expected "{", was ";" (Sass::SyntaxError)
```

When I remove the front matter, `sass` and `scss-lint` are able to parse and lint it, but then Jekyll doesn't recognize it as a Sass stylesheet. So I keep the front matter and push it to github, but then Hound CI reports these errors as fatal when linting.

I wasn't able to find any solution to this by searching the issues or googling, but is this a known problem? Is there a workaround, or is the fix to have scss-lint strip front matter before sending it to Sass for parsing?

sds:
See the section in the README on [preprocessing](https://github.com/brigade/scss-lint#preprocessing).

connorshea:
Just FYI if you have this problem and all your stylesheets have the same number of frontmatter variables you can use `preprocess_command: "sed '1,3d'"`, where 3 is the line upon which the "closing" `---` is found.

Thought this might be useful to someone in the future.

ortonomy:
https://github.com/brigade/scss-lint/issues/891

PR #870, opened 2016-11-02
Fix sed preprocess_command example

remy-actual:
Current sed command only removes first line of file. 
Edited command removes entire Front Matter block.

### DIFF
--- a/README.md
+++ b/README.md
@@ -493,7 +493,7 @@ For example, `preprocess_command: "cat"` specifies a simple no-op preprocessor
 (on Unix-like systems). `cat` simply writes the contents of STDIN back out to
 STDOUT. To preprocess SCSS files with
 [Jekyll front matter](http://jekyllrb.com/docs/assets/), you can use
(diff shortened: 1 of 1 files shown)


sds:
Thanks for the pull request, @remy-actual.

Your command does not work for me. Here's what I get when I run the following using the built-in version of `sed` that comes with macOS:

```bash
$ sed "1{/^---$/{:a N;/---$/bash --version;d}}" screen.scss
sed: 1: "1{/^---$/{:a N;/---$/ba ...": unexpected EOF (pending }'s)
```

Perhaps you are using a different version of `sed`? Ideally we'd have a command that works on both macOS and Linux. Thanks!

remy-actual:
Yes, I am using gnu-sed installed from homebrew with --with-default-names flag.  
However, sed command currently in the README doesn't work with either gnu-sed or Mac OS X's default FreeBSD sed. The first `---` line is removed, not the full front matter block. 
I've added a notice, as well as a little more info concerning the use case: Issue #787

sds:
Merged in 2615dfade9b741e678931abdac1598cce3058fb1. Thanks!

Issue #516, opened 2015-07-08
Use the first .scss-lint.yml configuration file found in the current working directory or one of its ancestors


webgefrickel:
I understand how the configuration files work at the moment (.scss-lint-files by --config flag >> CWD >> ~), and no merging. I just ran into an 'issue' in a very large project with different teams regarding the configuration of the linter - because everybody wanted to do sth. just a little bit different :/ (tabs vs. spaces etc.)
The solution then was, that every team got their own config-file, with their own customizations. This resulted in a lot of duplicated code in the .scss-lint.yaml-files (and lots of those files as well). Adding or changing rules in the future would thus be quite tedious, because it has to be done in lots of files...

I know that this is more a political issue and should be solved by just sticking to one default standard for all teams, but could you consider sth. like eslint's configuration cascade? http://eslint.org/docs/user-guide/configuring#configuration-cascading-and-hierarchy

The idea here: every file is being checked by using 'the closest' .scss-lint.yaml file - if none in the current dir, move to the parent and so on - and as a last resort ~/.scss-lint.yaml. This way I could configure a 'sane default', and then let the teams just add their custom rules on top.

Would this be possible at all, or do you even think that this would be a bad idea?

Anahkiasen:
> because everybody wanted to do sth. just a little bit different :/

Wouldn't go, like, against the purpose of having a linter in the first place?

webgefrickel:
I know this is more a political issue. But regarding different teams: inside the team everybody sticks to their rules it is not really against the purpose of a linter (and they quite like it) — but the rules between the different teams don't match. Political stuff like tabs vs. spaces aside e.g. one team wants to be 'more strict' than the others, disallowing @extend or extending from classes and making use of the new BemDepth-Linter. I know that those are edge-cases, just some food for thought :)

srawlins:
So that its not a breaking change for everyone, it could be implemented as a feature that requires `cascade: true` at the top-level of a .scss-lint.yaml, in order to merge with a deeper one.

sds:
`scss-lint` used to support "inheriting" configurations but removed it in 56cbbd2f68779fd3906c5c136483a2cfafc12a29. Pasting the commit message here because I think it's informative about why:

>    Remove inherit_from and per-directory configuration support
>
>    These two features were created very early in `scss-lint`'s history, and
>    were largely a result of trying to emulating the `rubocop` project's
>    configuration system.
>
>    Over time, it's become clear that the configuration system is the most
[...]

PR #878, opened 2016-12-29
Allow loading config file from plugin directory

barraq:
# Description

`plugin_directories` property can be used in `.scss-lint.yml` to load additional linters from a directory. However it does not allow loading custom `.scss-lint.yml` for such directory.

This PR allows loading a custom `.scss-lint.yml` from any directory specified in `plugin_directories` and merge it with your existing configuration. This is pretty handy to allow extending your existing configuration from other configuration coming from other non-gem package, e.g. npm, bower, etc.

# What was done

- [x] Update `linter_dir.rb` to add loading custom `.scss-lint.yml` functionality (based on `linter_gem`)
- [x] Update the spec accordingly

# Note

While `scss-linter` allows loading `.scss-lint.yml` file from other `gems` it sometimes happen that you need to extend your project config from another `.scss-lint.yml` that is not coming from a `gem` but from another directory or another non-gem package, like npm, bower, etc.

```
# .scss-lint.yml
plugin_directories: ['.scss-linters', 'node_modules/custom_package']
```

Might be related to https://github.com/brigade/scss-lint/issues/516

### DIFF
--- a/lib/scss_lint/plugins/linter_dir.rb
+++ b/lib/scss_lint/plugins/linter_dir.rb
@@ -6,11 +6,11 @@ module SCSSLint
 
       def initialize(dir)
         @dir = dir
-        @config = SCSSLint::Config.new({}) # Will always be empty
       end
 
       def load
         ruby_files.each { |file| require file }
+        @config = plugin_config
         self
       end
 
@@ -18,6 +18,29 @@ module SCSSLint
 
       def ruby_files
         Dir.glob(File.expand_path(File.join(@dir, '**', '*.rb')))
+      end
+
+      # Returns the {SCSSLint::Config} for this directory.
+      #
+      # This is intended to be merged with the configuration that loaded this
+      # plugin.
+      #
+      # @return [SCSSLint::Config]
+      def plugin_config
+        file = plugin_config_file
+
+        if File.exist?(file)
+          Config.load(file, merge_with_default: false)
+        else
+          Config.new({})
+        end
+      end
+
+      # Path of the configuration file to attempt to load for this directory.
+      #
+      # @return [String]
+      def plugin_config_file
+        File.join(@dir, Config::FILE_NAME)
       end
     end
   end
--- a/spec/scss_lint/plugins/linter_dir_spec.rb
+++ b/spec/scss_lint/plugins/linter_dir_spec.rb
@@ -4,18 +4,42 @@ describe SCSSLint::Plugins::LinterDir do
   let(:plugin_directory) { File.expand_path('../../fixtures/plugins', __FILE__) }
   let(:subject) { described_class.new(plugin_directory) }
 
-  describe '#config' do
-    it 'returns empty configuration' do
-      subject.config.should == SCSSLint::Config.new({})
-    end
-  end
-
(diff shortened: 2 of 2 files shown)


lib/scss_lint/plugins/linter_dir.rb
```diff
@@ -19,6 +19,29 @@ def load
...
+        else
+          Config.new({})
+        end
+      end
+
+      # Path of the configuration file to attempt to load for this directory.
+      #
+      # @return [String]
+      def plugin_config_file
+        File.join(@dir, Config::FILE_NAME)
```
trotzig: To reduce indirection, this line could probably be inlined directly in the `plugin_config` method (since you don't use it elsewhere).

lib/scss_lint/plugins/linter_dir.rb
```diff
@@ -19,6 +19,29 @@ def load
       def ruby_files
         Dir.glob(File.expand_path(File.join(@dir, '**', '*.rb')))
       end
+
+      # Returns the {SCSSLint::Config} for this directory.
+      #
+      # This is intended to be merged with the configuration that loaded this
+      # plugin.
```
trotzig: This comment confuses me a little since we use `merge_with_default: false` below. What is indented to merge, and where does that happen? I don't have a lot of context on how scss-lint configuration works, so excuse me if I'm missing something obvious.

barraq: Plugins are loaded in Config.rb (see https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/config.rb#L235 and https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/plugins.rb). Basically each plugin is loaded and brings its own configuration. Then all those configs are merged into one config and eventually (unless specified otherwise in the main .scss-lint.yml or cli option) merge with the default.yml config. Therefore, each plugin should avoid merging its config from default as it is done at the end of the process.

I borrowed the comment from https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/plugins/linter_gem.rb#L25 for consistency.

lib/scss_lint/plugins/linter_dir.rb
barraq: I tried to keep as close a PluginGem structure (https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/plugins/linter_gem.rb#L44) but yes it can be inlined; unless we want to test it (which we don't). I can remove it ;)

barraq:
Hi @trotzig thanks for taking time to review this PR 👍

lib/scss_lint/plugins/linter_dir.rb
trotzig: Thanks for the explanation, that makes sense.

trotzig:
Thanks @barraq! I'm waiting to hear back from a person with more context on the setup. I'll get back soon.

trotzig:
Thanks for the contribution @barraq! I'm going to include this in a 0.52.0 release which will be published as soon as I have permissions to push to rubygems.org.

barraq:
Awesome :D thanks

Issue #883, opened 2017-01-18
Update Smacss


Yimiprod:
Is it possible to update the `.smacss.txt` file to support new css attributes?
I've worked on it https://gist.github.com/Yimiprod/c9dadba643d54c56c660a80755b3deaf and looking for advice to update it.

By the way i don't find how to define a custom `.txt` file for scss-lint.

trotzig:
What are the new attributes you're looking to add? Can you file a PR instead of the gist? I'll be happy to review it.

Yimiprod:
Here's a PR https://github.com/brigade/scss-lint/pull/884

PR #884, opened 2017-01-18
update: smacss order

Yimiprod:
https://github.com/brigade/scss-lint/issues/883

### DIFF
--- a/data/property-sort-orders/smacss.txt
+++ b/data/property-sort-orders/smacss.txt
@@ -1,14 +1,24 @@
-# SMACSS Property Order
+# based on SMACSS Property Order
 # http://smacss.com/book/formatting
+
+# Heading
+
+content
+quotes
 
 # Box
 
 display
+visibility
+
 position
+z-index
 top
 right
 bottom
 left
+
+box-sizing
 
 flex
 flex-basis
@@ -26,7 +36,6 @@ order
 width
 min-width
 max-width
-
 height
 min-height
 max-height
@@ -45,7 +54,13 @@ padding-left
 
 float
 clear
+
+overflow
+overflow-x
+overflow-y
+
 clip
+zoom
 
 columns
 column-gap
@@ -55,16 +70,39 @@ column-span
 column-count
 column-width
 
+table-layout
+empty-cells
+caption-side
+border-spacing
+border-collapse
+list-style
+list-style-position
+list-style-type
+list-style-image
+
+# Animation
+
 transform
-transform-box
 transform-origin
 transform-style
+backface-visibility
+perspective
+perspective-origin
 
 transition
-transition-delay
-transition-duration
 transition-property
+transition-duration
 transition-timing-function
+transition-delay
+
+animation
+animation-name
+animation-duration
+animation-play-state
+animation-timing-function
+animation-delay
+animation-iteration-count
+animation-direction
 
 # Border
 
@@ -103,16 +141,24 @@ outline-offset
 outline-style
 outline-width
 
+stroke-width
+stroke-linecap
+stroke-dasharray
+stroke-dashoffset
+stroke
+
 # Background
 
+opacity
+
 background
-background-attachment
-background-clip
 background-color
 background-image
 background-repeat
 background-position
 background-size
+box-shadow
+fill
 
 # Text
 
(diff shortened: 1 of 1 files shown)


trotzig:
This is quite a big change to the order. Could you explain the reasoning a little? For instance, you've moved z-index from the "other" group into the "box" group.

Yimiprod:
I'm not sure it's the good way, but since there's no reference about ordering this rules, in particular the **others**, it's based on way people tend to write thoses rules.

Some of those choices certainly deserve more discution and more people to talk about those orders.

**Why moving _other_ properties :**
- **content** on first position, almost only used in pseudo-elements, that way it's the first properties for those elements and never forgot to put that propertie.
- **quotes** just after, same reasoning as **content**, since it's used in specific element.
- **visibility**, since it's a display propertie, certainly deserve more reflection
- **z-index**, some people tend to like it at last position, it's arguable, but it's a position propertie
- **overflow**, missing x and y, it was let in other, but it certainly belong to box model
- **table-layout** was missing, it's a box model
- **animations** probably at the wrong place in this order, i'm not sure about that point, but they were missing
- **stroke** was missing in border properties, it's the border of an svg element
- **opacity** was left in other, moved it in background since it doesn't change box model but doesn't change the display, it subjectif
- **box-shadow** should belong with background, it change the way the box appear
- **fill** was missing, it's the background of an svg element
- **text** many font properties was missing, ordered mainly by name.

I hope it's clear, i've added some new properties in other, not knowing where to add thems, maybe deserve a User interaction section.

I'm open to any changes, and if there's a place to talk about properties ordering, it would be good to have the link.

trotzig:
Thanks. We're not using this rule ourselves at Brigade, so it would be good to have some outside feedback by people who use it. 

I'm pinging a few people who have made contributions to smacss in the past, to see if they have any input: @danny-andrews, @tobiashm, @bkzl, @ingdir, @tgecho.

bkzl:
I like the changes proposed by @Yimiprod and agree with all of the above. +1

data/property-sort-orders/smacss.txt
```diff
@@ -1,15 +1,23 @@
 # SMACSS Property Order
 # http://smacss.com/book/formatting
 
+content
```
tobiashm: I think it would be a good idea to precede this with a comment about these properties being outside of the groups listed at https://smacss.com/book/formatting#grouping (sine the following rules are grouped by a comment “heading”)

data/property-sort-orders/smacss.txt
```diff
@@ -121,41 +167,66 @@ color
 font
 font-family
 font-size
+font-size-adjust
+font-stretch
+font-effect
+font-smooth
 font-smoothing
+-moz-osx-font-smoothing
+-webkit-font-smoothing
```
tobiashm: There're no other vendor prefixed rules in these lists, so I don't think it's a good idea to start adding these here.

tobiashm:
I think this looks pretty good. There're not many specifics listed at https://smacss.com/book/formatting#grouping except for the general grouping principles. So maybe it would be a good idea to add some comments about this being _based on_ SMACSS, with some additional principles for property orders, just so that others don't take this to be an official expression of Jonathan Snooks intent.

data/property-sort-orders/smacss.txt
Yimiprod: Will do. Thanks

data/property-sort-orders/smacss.txt
Yimiprod: Well to be fair, i added them because those two properties are mandatory when doing a custom font and it's commonly used.
But i think i'll remove them, since it's not standard at the moment.

Yimiprod:
Maybe i should ask him what he think about it?

Yimiprod:
@snookca look not active on github, since i don't want to lure people, i'll ask him in a mail.
If no i get no anwser i'll add the comments.

Thank for the reviews.

data/property-sort-orders/smacss.txt
trotzig: Good call. I suggest you look into using something like [autoprefixer](https://github.com/postcss/autoprefixer) to avoid having to use vendor-prefixes.

Yimiprod: I'm thinking about deleting **font-smooth** and **font-smoothing** since they aren't standard.

CeliaLfd:
Following the proposal of @Yimiprod, it would be a good idea to think about the properties `grid-layout` that will be supported soon.

Yimiprod:
Got a response by mail from @snookca where he seem to aggree
> Happy to hear the book has been helpful!
> I took a look at that PR. No specific objections with what you’ve outlined. All makes sense to me. :) 
image

trotzig:
The Travis failure is for the jruby build, most likely not related to your changes. Thanks @Yimiprod for contributing to this project!

PR #892, opened 2017-03-14
Add missing transform-origin property to concentric order

dFegnoux:


trotzig:
Thanks!

psa-dfegnoux:
I saw that property missing in recess order too but i'm not familiar with it sorry ;)

PR #895, opened 2017-03-22
Allows the ignoring of variables being reported in UrlFormatter

ethikz:
Currently only ignores data but now also ignores `${`

### DIFF
--- a/lib/scss_lint/linter/url_format.rb
+++ b/lib/scss_lint/linter/url_format.rb
@@ -43,7 +43,7 @@ module SCSSLint
     end
 
     def check_url(url, node)
-      return if url.start_with?('data:')
+      return if url.start_with?('data:') || url.include?('${')
       uri = URI(url)
 
       if uri.scheme || uri.host
--- a/spec/scss_lint/linter/url_format_spec.rb
+++ b/spec/scss_lint/linter/url_format_spec.rb
@@ -33,7 +33,11 @@ describe SCSSLint::Linter::UrlFormat do
     end
 
     context 'when URL contains a variable' do
-      let(:url) { '${' }
+      let(:scss) { <<-SCSS }
+        .block {
+          background: url('${url}');
+        }
+      SCSS
 
       it { should_not report_lint }
     end
--- a/spec/scss_lint/linter/url_format_spec.rb
+++ b/spec/scss_lint/linter/url_format_spec.rb
@@ -33,7 +33,11 @@ describe SCSSLint::Linter::UrlFormat do
     end
 
     context 'when URL contains a variable' do
-      let(:url) { '${url}' }
+      let(:scss) { <<-SCSS }
+        .block {
+          background: url('${url}');
+        }
+      SCSS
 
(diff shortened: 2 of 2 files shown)


spec/scss_lint/linter/url_format_spec.rb
```diff
@@ -31,6 +31,12 @@
 
       it { should_not report_lint }
     end
+
+    context 'when URL contains a variable' do
+      let(:url) { '${' }
```
trotzig: This would be better if it had the full string, e.g. `'${some-var}'`. I know that your code doesn't care, but I could see someone changing the check to a more involved regex matching the whole variable inject.

ethikz: Well the idea is to not fail if the url has an sort of variable match.  I mean I can change it so it allows `#{` or `{{` but isn't the idea that leaving it open like that it doesn't matter?

or should it be `{ '${url}' }?

ethikz:
Any clue why this keeps failing?  I only added the one thing?

spec/scss_lint/linter/url_format_spec.rb
ethikz: @trotzig Because in some of our cases we use variables because the CSS goes into a CMS that replaces those variables but while we are prototyping it never passes linter because of that

ethikz:
@trotzig Thanks!  With what I did should your above referenced code be ignored?  Right now it reports as an error when it should be ignored because it is a variable.  The implementation I put in I thought would not report that as an error.

`url('/response/${siteId}/artifact/${campaignAndToolPath}/assets/fonts/roboto-light-webfont.woff2')`

Before the change this was reported as an error which it shouldn't be because it is a valid URL just with a variable that the CMS replaces.

trotzig:
Interesting, I should have checked my assumptions! I didn't know `$` is a [reserved URI character](https://en.wikipedia.org/wiki/Percent-encoding#Types_of_URI_characters), and that it has to be escaped. 

```rb
irb(main):003:0> URI.parse('http://domain.com/${foobar}.jpq')
URI::InvalidURIError: bad URI(is not URI?): http://domain.com/${foobar}.jpq
```

Thanks for making me double-check this!

ethikz:
Ahh good catch!

Is there anything I need to do? to fix it?

ethikz:
Ok awesome!!  Glad I could help and must say I had help as well from Stack Overflow since as stated I'm not a Ruby genius, I know a little here and there.

When will this be released so I can update?

trotzig:
This is now released in [v0.53.0](https://github.com/brigade/scss-lint/releases/tag/v0.53.0). 🍾

lib/scss_lint/linter/url_format.rb
```diff
@@ -43,7 +43,7 @@ def url_string?(arg)
     end
 
     def check_url(url, node)
-      return if url.start_with?('data:')
```
ethikz: what ?

PR #908, opened 2017-07-05
Add support for the ::ng-deep pseudo-element

nex3:
Angular has added the `::ng-deep` pseudo-element in angular/angular@b754e600e30476a22e29019d89df71dd9b7fcbd5 as a CSS-compliant replacement for `/deep/` and `>>>`. This avoids having the linter complain about this element.

### DIFF
--- a/data/pseudo-elements.txt
+++ b/data/pseudo-elements.txt
@@ -110,6 +110,7 @@ before
 content
 first-letter
 first-line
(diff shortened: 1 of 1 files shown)


sds:
Thanks for the PR, @nex3! And doubly so for the incredible work you've done on Sass over the years. 🙇

PR #909, opened 2017-07-31
Update example for JSON formatter in the README

pocke:
The JSON formatter examples does not have `column`, `length` and `litner` fields in the README. However, those fields are included actually.

For example:

```bash
$ scss-lint test.scss --format=JSON
{
  "test.scss": [
    {
      "line": 2,
      "column": 1,
      "length": 1,
      "severity": "warning",
      "reason": "Declaration should be terminated by a semicolon",
      "linter": "TrailingSemicolon"
    },
    {
      "line": 2,
      "column": 10,
      "length": 4,
      "severity": "warning",
      "reason": "Color literals like `#111` should only be used in variable declarations; they should be referred to via variable everywhere else.",
      "linter": "ColorVariable"
    }
  ]
}
```

This change adds those fields to the example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,10 +308,10 @@ Outputs JSON with filenames and an array of issue objects.
 ```json
 {
   "test.css": [
-    {"line": 2, "severity": "warning", "reason": "Prefer single quoted strings"},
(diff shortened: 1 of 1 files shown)


sds:
Thanks for helping keep our documentation up to date, @pocke!

PR #913, opened 2017-09-22
Classic BEM selector format

### LINKED ISSUE #859: Trying to use strict BEM but scss-lint complains
.block-name__element-name_modifier-name doesnt work ..i am bit confused 

.block-name__element-name -> works
.block-name_modifier-name -> works 
.block-name__element-name_modifier-name -> doesnt 

am i missing something ? thanks for your time

aar0nr:
The official BEM syntax is:

* Block: ```.block-name```
* Element: ```.block-name__elem-name```
* Modifier (boolean): ```.block-name_mod-name```
* Modifier (key/value): ```.block-name_mod-name_mod-val```

Non-standard styles, but accepted are:

* Two dashes: ```.block-name__elem-name--mod-name```
* Camel case: ```.MyBlock__SomeElem_modName_modVal```
* Sans underscore: ```.blockName-elemName--modName--modVal```

The ```strict_BEM``` convention for the ```SelectorFormat``` linter is supposed to implement the official BEM style from my understanding, but it doesn't. For example, it reports ```.block_modifier``` and ```.block__element_modifier```  invalid, but are valid classic BEM styles.

I've added a new convention which I call ```classic_BEM``` to adhere to the official BEM style. I use the term *classic* because that's how the creators [refer to it](https://en.bem.info/methodology/naming-convention/#which-style-to-choose).

Instead of removing or fixing the ```strict_BEM``` convention (which many people may be using), I think it should be deprecated since it's confusing which style it's actually adhering to.

Fixes #859

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## master (unreleased)
 
 * Add `ng-deep` to the list recognized by the `PseudoElement` linter
+* Add `classic_BEM` convention for `SelectorFormat`
 
 ## 0.54.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 * Add `ng-deep` to the list recognized by the `PseudoElement` linter
 * Add `classic_BEM` convention for `SelectorFormat`
+* Deprecate `strict_BEM` convention for `SelectorFormat`.
+  Use `classic_BEM` instead.
 
 ## 0.54.0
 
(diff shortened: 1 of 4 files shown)


aar0nr:
If you decide you want to deprecate ```strict_BEM``` I can create a new issue for the 1.0 milestone to remove it.

sds:
Thanks for the pull request, @aar0nr. I don't think any of this project's maintainers are intimately familiar with the BEM methodology, so we'll take you at your word that this is what the original BEM creators intended.

Thank you for writing tests to ensure that this behavior is specified in code.

Issue #877, opened 2016-12-16
Add support for Sass 3.5


mprins:
I'm looking at running scss_lint with the Sass 3.5 pre-release [in my branch](https://github.com/mprins/scss-lint/tree/sass-3.5) and the only issue I seem to be hitting is som failing testcases due to:

``` ruby
     NoMethodError:
       undefined method `to_sass' for ["1px solid #000"]:Array
       Did you mean?  to_s
```
(see [travis builds](https://travis-ci.org/mprins/scss-lint/builds/184488559))

Looking at the [Sass API docs](http://sass-lang.com/documentation/Sass/Script/Value/Base.html#to_s-instance_method) it seems the "proper name" is `to_s` but I might be overlookig something.
Is there an interest to have this adjusted (`to_sass` -> `to_s`)? and have things work on 3.5?

As a side note, but unrelated to this it seems that while the native rubies cause a failing build the jruby is green :confused:

sds:
Hey @mprins, thanks for looking into running SCSS-Lint on Sass 3.5.

Looks like Travis lied to you, as the [JRuby build definitely failed but reported as a pass](https://travis-ci.org/mprins/scss-lint/jobs/184488564). Looks like this is because while the RSpec tests failed the second step of our tests passed because we skip the Overcommit pre-commit checks for JRuby (since they aren't useful and usually fail).

In any case, would love a pull request that changes `to_sass` -> `to_s` if it results in all the tests passing. We'll probably want to modify our Travis build matrix to test on both Sass 3.4 and 3.5, if possible.

mprins:
did a quick search-replace of `to_sass` -> `to_s` and ended up wit 60 or so test failures when using 3.5..; probably should start off with doing that code change first and try that with 3.4.x

tagliala:
Hi,

I did an experiment in this commit: https://github.com/tagliala/scss-lint/commit/5ab8376cd335efa795946250dd1e2e5532664b74

All specs are green but I'm not confident of the solution. Please read commit's message to understand what I changed and why

sds:
Thanks for digging into this, @tagliala. Your general approach seems correct, but will not work in some cases. It seems `Sass::Tree::PropNode#value` now returns an array to represent interpolation in property values (see https://github.com/sass/sass/commit/8247d55b0).

[...]

Issue #750, opened 2016-03-07
Stop monkey-patching `Sass::Tree::Node#children`


sds:
While it is documented that SCSS-Lint monkey patches the Sass parse tree and that developers should not `require` it with their regular application code, it continues to be a source of confusion and frustration.

I think we can get around this by changing the `Linter` class to not inherit from `Sass::Tree::Visitors::Base` and instead define our own methods for traversal, saving us from having to monkey-patch the `children` method of the `Sass::Tree::Node` class and its descendants. It's low-level enough that it shouldn't require us to touch any linter tests, and they'll quickly give us a heads up if we've horribly broken something.

Issue #278, opened 2014-10-23
Undefined method `bubbles?`


mmozuras:
After adding `scss-lint` to our project Gemfile test/development group (to be primarily used with [Pronto](https://github.com/mmozuras/pronto)), we got an error during assets precompile:

`undefined method invalid_parent_method_name`

After adding the necessary methods ([pull request here](https://github.com/causes/scss-lint/pull/277)), we now see:

`undefined method bubbles? for #31abc2`

Any tips on how to proceed? Or can't `scss-lint` be used like that, because it messes with `sass` too much?

agross:
How would you go about this with guard?

I use guard with nanoc to compile static HTML (and SCSS to CSS). I never require scss_lint, just `bundle exec guard`. Whenever nanoc compilation kicks in I get:

```
NoMethodError: undefined method `bubbles?' for 50px:Sass::Script::Tree::Literal
```

Adding `require: false` to my scss_lint-related gems (+ checkstyle, guard-scss_lint) doesn't remove the error.

sds:
The moment you run `require 'scss_lint'` in a Ruby process (which is what happens when you `bundle exec rake scss_lint...`) you're hosed&mdash;SCSS-Lint has monkey-patched the parse tree which causes any other Sass library calls to likely fail.

I've opened #750 to track fixing this issue.

agross:
@sds Thanks! :+1:

PR #923, opened 2017-10-28
Release sass dependency

printercu:
Hi! I've got this error:
```
    bootstrap (= 4.0.0.beta2.1) was resolved to 4.0.0.beta2.1, which depends on
      sass (>= 3.5.2)

    scss_lint was resolved to 0.38.0, which depends on
      sass (~> 3.4.1)
```

It should be ok not to use `~>`, as only `sass@4.x.x` may introduce breaking changes.

### DIFF
--- a/lib/scss_lint.rb
+++ b/lib/scss_lint.rb
@@ -31,3 +31,28 @@ require 'scss_lint/reporter'
 Dir[File.expand_path('scss_lint/reporter/**/*.rb', File.dirname(__FILE__))].sort.each do |file|
   require file
 end
+
+# Sass::Tree::PropNode#value returns an array in new versions. Here is note from docs:
+#
+#   For most properties, this will just contain a single Node. However, for
+#   CSS variables, it will contain multiple strings and nodes representing
+#   interpolation.
+#
+# This monkey-patch allows us use scss-lint with new versions of sass.
+Array.class_eval do
+  def to_sass_value
+    raise 'This proxy does not support css variables' unless size == 1
+    first
+  end
+
+  def to_sass
+    to_sass_value.to_sass
+  end
+end
+
+Sass::Script::Tree::Node.class_eval do
+  # We use #to_sass_value to keep it compatible to sass < 3.5.
+  def to_sass_value
+    self
+  end
+end
--- a/lib/scss_lint/linter/duplicate_property.rb
+++ b/lib/scss_lint/linter/duplicate_property.rb
@@ -54,6 +54,7 @@ module SCSSLint
     end
 
     def value_as_string(value)
+      value = value.to_sass_value
       case value
       when Sass::Script::Funcall
         value.name
--- a/lib/scss_lint/linter/property_units.rb
+++ b/lib/scss_lint/linter/property_units.rb
@@ -32,8 +32,9 @@ module SCSSLint
         property = "#{@nested_under}-#{property}"
       end
 
-      if node.value.respond_to?(:value)
-        node.value.value.to_s.scan(NUMBER_WITH_UNITS_REGEX).each do |matches|
+      value = node.value.to_sass_value
+      if value.respond_to?(:value)
+        value.value.to_s.scan(NUMBER_WITH_UNITS_REGEX).each do |matches|
           is_quoted_value = !matches[0].nil?
           next if is_quoted_value
           units = matches[1]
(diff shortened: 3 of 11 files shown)


printercu:
Seems like not everything so clear :( I've got error with this configuration:

```
SCSSLint::Linter::BorderZero raised unexpected error linting file app/assets/stylesheets/shared/_tables.scss: 'undefined method `to_sass' for ["none"]:Array
Did you mean?  to_s'
```

tagliala:
@printercu please take a look at #877, https://github.com/brigade/scss-lint/issues/877#issuecomment-339966451 and https://github.com/tagliala/scss-lint/commit/ac247889acec6841c4aef81655d8ecc83a373f4d

printercu:
@tagliala thanks! I've checked your commit, it helped to understand. I've used slightly different approach: I added `#to_sass_value` to Array and sass Node. This way it will work in both 3.4 and 3.5 sass versions.

I think we should better add appraisal to test against different sass versions.

tagliala:
>  it looks more like sass issue

@printercu are you able to provide a spec which fails in Sass only? I'm still having troubles in understand what really happens when visiting the tree

printercu:
I even haven't thought of it before, just saw backtrace full of `/gems/sass-3.5.3/lib/sass/tree/visitors/` and thought that it was sass. Now I see that it happens only after `scss_lint/sass/tree` is required. Everything is much more complicated than it seemed to be :)

I'm not sure if i'll be able to try to fix it soon. Maybe i'll really try to migrate to other libs or use this gem until i face an error. For now, with this patch it seems to work for our small codebase.

sds:
Thanks for taking time to dig into this issue, @printercu and @tagliala. I really appreciate the effort and understand it's frustrating that this doesn't seem to be working as hoped.

Given that the problem seems to be caused when we require `scss_lint/sass/tree`, my guess is that the nasty monkey-patching we're doing is causing issues. I figured this would one day come to bite us.

#750 suggests a potential workaround to this problem, which is to not have `Linter` inherit from `Sass::Tree::Visitors::Base`, but define its own methods for traversal.

As you can see, that issue has been around for a while. I no longer work at @brigade and I don't use `scss-lint` at my current employer, so any work I do on this project is entirely out of free time. I think it's possible for someone to hack a solution to #750 in a day, but there may be unknown challenges with that path. I would suggest someone solve #750 independently of everything here (getting all existing tests passing), and getting that merged. 

Once we stop monkey patching we can proceed to tackle the challenge here. Hopefully the work will pay off and we won't hit this strange issue where we attempt to visit a `String`.

However, as things currently stand I'm not comfortable with merging something that breaks on Twitter Bootstrap's codebase, especially when the reproducible test case is so simple. Open to hearing your thoughts.

tagliala:
@sds thanks for your time and for your detailed explanation

I think I need way more time of one day to implement #750, because I need to understand what's behind SCSS-Lint.

Also, I was experiencing #278 in Rails 5.1's system tests but I didn't have time to investigate. Now I realize that it is caused bt scss-lint's monkey patch to Sass, so I think that #750 is a must-have

```
$ rake scss_lint assets:clobber tmp:clear test:system
scss-lint found no lints
Run options: --seed 29292

# Running:

Puma starting in single mode...
* Version 3.10.0 (ruby 2.4.2-p198), codename: Russell's Teapot
* Min threads: 0, max threads: 1
* Environment: test
* Listening on tcp://0.0.0.0:62911
Use Ctrl-C to stop
2016-10-10 00:00:00 +0200: Rack app error handling request { GET /calendar }
#<ActionView::Template::Error: undefined method `bubbles?' for null:Sass::Script::Tree::Literal>
~/.rvm/gems/ruby-2.4.2/gems/sass-3.4.25/lib/sass/tree/visitors/check_nesting.rb:157:in `trans
```

without scss-lint
```
$ rake scss_lint assets:clobber tmp:clear test:system
scss-lint found no lints
Run options: --seed 29292

# Running:

Puma starting in single mode...
* Version 3.10.0 (ruby 2.4.2-p198), codename: Russell's Teapot
* Min threads: 0, max threads: 1
* Environment: test
* Listening on tcp://0.0.0.0:62911
Use Ctrl-C to stop
2016-10-10 00:00:00 +0200: Rack app error handling request { GET /calendar }
#<ActionView::Template::Error: undefined method `bubbles?' for null:Sass::Script::Tree::Literal>
~/.rvm/gems/ruby-2.4.2/gems/sass-3.4.25/lib/sass/tree/visitors/check_nesting.rb:157:in `trans
```

Of course, by omitting scss_lint or by changing order of tests this issue doesn't occurr

sds:
Closing in favor of #927. Thanks for your help here!

PR #925, opened 2017-11-01
Make all lint to have `linter` attribute

pocke:
Problem
===

Currently SCSS-Lint reports syntax and encoding issues without `linter` attribute.
For example:

`test.scss`

```scss
{
```

`valid.scss`

```scss
a {
  border: none;
}
```

```bash
$ scss-lint --format JSON
{
  "test.scss": [
    {
      # This issue does not have `linter` attribute.
      "line": 1,
      "column": 1,
      "length": 1,
      "severity": "error",
      "reason": "Syntax Error: Invalid CSS after \"\": expected selector or at-rule, was \"{\""
    }
  ],
  "valid.scss": [
    {
      "line": 2,
      "column": 3,
      "length": 12,
      "severity": "warning",
      "reason": "`border: 0` is preferred over `border: none`",
      "linter": "BorderZero" # This issue has `linter` attribute.
    }
  ]
}
```

This behaviour is difficult to handle. We develop [SideCI](https://sideci.com) that is a automated code review service. SideCI uses SCSS-Lint, and we expect that all issues have a issue type, that is called "linter" in SCSS-Lint. But SCSS-Lint's syntax issue does not have "linter". So our program crashes.

Solution
====

Make all lint to have `linter` attribute.
In other words, I added `Linter::Syntax` and `Linter::Encoding`class, and use these classes instead of nil, and removed `lint.linter` existence check.

### DIFF
--- a/lib/scss_lint/linter/encoding.rb
+++ b/lib/scss_lint/linter/encoding.rb
@@ -1,0 +1,4 @@
+module SCSSLint
+  class Linter::Encoding < Linter
+  end
+end
--- a/lib/scss_lint/linter/syntax.rb
+++ b/lib/scss_lint/linter/syntax.rb
@@ -1,0 +1,4 @@
+module SCSSLint
+  class Linter::Syntax < Linter
+  end
+end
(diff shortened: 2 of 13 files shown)


sds:
Thanks for the pull request, @pocke. 

Rebased and merged in bdde989d.

pocke:
Thank you!

PR #927, opened 2017-11-04
Support Sass 3.5

### LINKED ISSUE #877: Add support for Sass 3.5

tagliala:
This is not the PR we deserve, but it's the one we need right now. 

Sass 3.5 is supported, custom properties aren't: all specs are green (included the new one which deals with custom properties), and it does not fail against Bootstrap v4

Also:
- Update development gems
- Test against latest Ruby versions
- Fix all new (0.50.0) RuboCop offences

Fix #877

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,5 +1,6 @@
 AllCops:
-  TargetRubyVersion: 1.9 # Oldest version of Ruby we support
+  DisplayCopNames: true
+  TargetRubyVersion: 2.0 # Oldest version of Ruby we support
 
 AccessModifierIndentation:
   EnforcedStyle: outdent
@@ -24,9 +25,6 @@ Documentation:
   Exclude:
     - 'spec/scss_lint/**/*'
     - 'spec/scss_lint/linter/**/*'
-
-Encoding:
-  EnforcedStyle: when_needed
 
 Eval:
   Exclude:
@@ -66,10 +64,10 @@ Layout/MultilineOperationIndentation:
   Enabled: false
 
 LineLength:
-  Max: 100
+  Max: 102
 
-# Reports false positives for `CONSTANT % ['value', 'value']` (RuboCop 0.33.0)
-Lint/FormatParameterMismatch:
+# Reports false positives for `'.one #{$interpolated-string} .two .three {}'` (RuboCop v0.50.0)
+Lint/InterpolationCheck:
   Enabled: false
 
 MethodLength:
@@ -103,9 +101,6 @@ PercentLiteralDelimiters:
 PredicateName:
   Enabled: false
 
-SignalException:
-  Enabled: false
-
 # Forcing a particular name (e.g. |a, e|) for inject methods prevents you from
 # choosing intention-revealing names.
 SingleLineBlockParams:
@@ -122,7 +117,4 @@ Style/TrailingCommaInArguments:
   Enabled: false
 
 Style/TrailingCommaInLiteral:
-  Enabled: false
-
-UselessAccessModifier:
   Enabled: false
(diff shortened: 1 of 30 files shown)


sds:
👏 🙇

tagliala:
@sds 

I want to submit a new PR that:

- Move development dependencies to the proper place (gemspec). Take a look [here](https://gemnasium.com/brigade/scss-lint): having them in the Gemfile lists them as runtime dependencies
- Drop Ruby 2.0 support. This doesn't mean that Ruby 2.0 will not work anymore, but Ruby 2.1 will be required to develop the gem and we will exclude tests on Ruby 2.0, so we can also use RuboCop 0.51.0
- Use `coveralls_reborn`, my (temporary I hope) fork of `coveralls`, which seems unmaintained.

What do you think?

sds:
Hey @tagliala, thanks for offering to move forward on some of these issues.

1. The reason we explicitly declared the RuboCop gem in the `Gemfile` was to avoid a warning about duplicate dependencies. See the commit message in 98655a735 for details. At its core, if another project includes SCSS-Lint as a development dependency, we don't want to force them to use the RuboCop version specified in the gemspec, since it's pinned to a specific version. If this isn't a practical issue, I'm open to suggestions here (is this behavior no longer the case?)

2. I'm fine with dropping Ruby 2.0 support, as it's been [EOLed since February 2016](https://www.ruby-lang.org/en/news/2016/02/24/support-plan-of-ruby-2-0-0-and-2-1/). I know your suggestion is to simply remove it from the Travis build matrix, but at that point we're not ensuring we support it, so I'd rather just drop support.

3. Is there a summary of what exactly has changed with `coveralls_reborn`?

Thanks!

tagliala:
Hi,

1. Got it, 👍 
2. Perfect, will do a separate PR just for this
3. It's not a big deal. It comes with updated development dependencies (mainly SimpleCov). I've submitted a couple of PRs in their repo three months ago, but it seems kinda unmaintained.
Ref: [diff](https://github.com/lemurheavy/coveralls-ruby/compare/v0.8.21...tagliala:v0.9.0?expand=1), [comment](https://github.com/lemurheavy/coveralls-ruby/pull/125#issue-249994014)

sds:
Thanks for linking to the diff for `coveralls_reborn`. The update to SimpleCov doesn't seem like it would benefit the SCSS-Lint project just yet (but if I'm missing something let me know) as none of the [changes](https://github.com/colszowka/simplecov/blob/v0.15.1/CHANGELOG.md) seemed to be something that would affect this project.

If you find yourself down the road having an issue with the coverage information reported by SimpleCov, then I would be open to merging `coveralls_reborn`. My hope is that your PR eventually gets merged into the main project soon.

tagliala:
> Perfect, will do a separate PR just for this (after RuboCop releases a new version, because it has a lot of issues right now)

RuboCop released a new version, but it has a lot of false positives, so I will keep waiting.

Moreover, Sass 3.5.4 has an issue (fixed in the stable branch) and a lot of tests with this gem fail  sass/sass#2423, so I will wait for a new release of Sass, too

PR #932, opened 2017-12-29
Add grid properties to SMACSS property order

tobiashm:
Add next to flex box properties since they are related as layout properties.
Positioned before based on the idea that you're more likely to have flex box inside grid than the other way around.


sds:
Thanks @tobiashm.

Rebased and merged in 3dd8055b3a0e3d41c28acb8f0da2ba6901d37c27.

PR #933, opened 2018-01-05
CI against Ruby 2.5

pocke:


sds:
Implemented in #935. Thanks!

PR #935, opened 2018-01-08
Drop Ruby 2.0

tagliala:
Also:

- Update Sass dependency to ~> 3.5.5 (sass/sass#2423) (#934)
- Update RuboCop
- Test against latest Ruby versions (#933)


sds:
Thanks @tagliala. Was there a specific reason we dropped support for Ruby 2.0? Just want to make sure we record that reason in the change log.

pocke:
Thank you for your pull-request!

PR #937, opened 2018-01-12
Do not warn about colors in lists in variables

### LINKED ISSUE #936: ColorVariable incorrectly warns on colors in lists
If I were to define a list, like a shadow:

```scss
$my-shadow: 2px 2px 2px rgba(0, 0, 0, 0.2);
$my-other-shadow: 2px 2px 2px #f00;
```

I currently get a ColorVariable warning for each line. ☹️

srawlins:
Fixes #936

### DIFF
--- a/lib/scss_lint/linter/color_variable.rb
+++ b/lib/scss_lint/linter/color_variable.rb
@@ -55,11 +55,13 @@ module SCSSLint
     def in_variable_declaration?(node)
       parent = node.node_parent
       parent.is_a?(Sass::Script::Tree::Literal) &&
-        parent.node_parent.is_a?(Sass::Tree::VariableNode)
+        (parent.node_parent.is_a?(Sass::Tree::VariableNode) ||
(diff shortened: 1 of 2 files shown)


lib/scss_lint/linter/color_variable.rb
```diff
@@ -55,11 +55,13 @@ def literal_string?(script_string)
     def in_variable_declaration?(node)
       parent = node.node_parent
       parent.is_a?(Sass::Script::Tree::Literal) &&
-        parent.node_parent.is_a?(Sass::Tree::VariableNode)
+        (parent.node_parent.is_a?(Sass::Tree::VariableNode) ||
+         parent.node_parent.node_parent.is_a?(Sass::Tree::VariableNode))
```
srawlins: I didn't see any guards elsewhere in this file for `parent_node` being nil. Should we sprinkle in some nil checks?

sds: There will always be a parent (at the very least, a `Sass::Tree::RootNode`).

PR #939, opened 2018-01-23
Add a missing pseudo element

RKushnir:
Proof that such an element exists 🙂 — https://github.com/WebKit/webkit/blob/b6e2e7a0f5a83f4205a21ac484fa4220a17b12cd/Source/WebCore/css/html.css#L576


trotzig:
Thanks!

PR #940, opened 2018-02-02
Add `at_least_one_space_or_newline` option to SpaceAfterPropertyColon

elstgav:
This rule would allow for the following:
```scss
height: 100px;
width:  100px;  // <-- multiple spaces

box-shadow:     // <-- newline
  inset 1px 1px 0 blue,
  0 0 5px red;
```

Currently, the `one_space_or_newline` option would complain about multiple spaces, and `at_least_one_space` would complain about the newline.


sds:
Thanks @elstgav!

PR #946, opened 2018-03-28
Add overscroll-behavior to the list of allowed properties.

zacharydenton:
See https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior for more information.


sds:
Thanks @zacharydenton!

PR #968, opened 2019-03-18
Add newest ruby versions for CI. And cutoff out of official support ruby versions.

nekomaho:
Hi! 

This gem does not seem to support yet a newer version of rubies.
This PR adds settings to CI to test the new versions of ruby.

And I remove old ruby versions that are already excluded from official support.

see also: https://www.ruby-lang.org/en/downloads/branches/

### About disabled rules
By removing the old ruby version, the following rubocop rules now detect existing source code.
As for these, in order to deviate from the purpose of PR, I add to the setting of exclusion for rubocop.
 - Style/FrozenStringLiteralComment
 - Style/SafeNavigation
 - Style/RedundantFreeze

If you want to enable those settings, I will make another PR.

thanks.


.travis.yml
```diff
@@ -11,11 +11,11 @@ env:
...
-  - 2.1.10
-  - 2.2.9
-  - 2.3.6
-  - 2.4.3
-  - 2.5.0
+  - 2.3.8
+  - 2.4.5
+  - 2.5.5
+  - 2.6.2
+  - ruby-head
```
sds: If you are adding a potentially unstable build, can please mark failures as ignored? We don't want broken builds because we're testing on unstable software.

nekomaho: Sorry for the late response...
I rethink about adding unstable build to CI, actually Adding unstable build seem makes no sense.
So I'll delete `ruby-head`.

.travis.yml
```diff
@@ -11,11 +11,10 @@ env:
...
-  - 2.1.10
-  - 2.2.9
-  - 2.3.6
-  - 2.4.3
-  - 2.5.0
+  - 2.3.8
+  - 2.4.5
+  - 2.5.5
+  - 2.6.2
   - jruby-9.1.15.0
```
mprins: Why not update JRuby as well?

nekomaho: I'm not familiar with JRuby...
I'll add 9.2.6.0 that newest version of JRuby.

sds:
Thanks @nekomaho!

PR #978, opened 2019-07-31
Add overflow-wrap

brookjordan:
`overflow-wrap` is the standard version of `word-wrap` and so should appear just after it.

### DIFF
--- a/data/property-sort-orders/smacss.txt
+++ b/data/property-sort-orders/smacss.txt
@@ -222,6 +222,7 @@ text-shadow
 text-transform
 text-wrap
 word-wrap
+overflow-wrap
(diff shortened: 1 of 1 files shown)


sds:
Test was failing for unrelated reason, which was fixed in 5739c3a150e5ab03996f041b1bb17f311ec37444.

data/property-sort-orders/smacss.txt
```diff
@@ -222,6 +222,7 @@ text-shadow
 text-transform
 text-wrap
 word-wrap
+overflow-wrap
```
brookjordan: ❤️

Issue #649, opened 2015-12-02
Add "scss-lint:disable-line"


callumlocke:
Disabling a rule for a single line is a bigger deal than it should be:

```scss
@import 'food';

// scss-lint:disable ImportPath
@import 'normalize-scss/_normalize.scss';
// scss-lint:enable ImportPath

body {
```

ESLint [has a nicer way of doing it](http://eslint.org/docs/user-guide/configuring), which I think scss-lint should steal:

```scss
@import 'food';
@import 'normalize-scss/_normalize.scss'; // scss-lint:disable-line ImportPath

body {
```

sds:
Hey @callumlocke,

You can already disable a single line by using a trailing comment.

```
@import 'normalize-scss/_normalize.scss'; // scss-lint:disable ImportPath
```

bbugh:
The README doesn't specify the single line format, and I found this via a web search.

BenLeggiero:
@bbugh Fixed! 😄

PR #982, opened 2020-01-20
Added single-line disable to README

BenLeggiero:
This was a hidden feature; as far as I could tell it was undocumented, but in the comments for #649, it was revealed that it is indeed an intentional feature. So, I added it to the README! 😄


sds:
Thanks for catching!

PR #986, opened 2020-03-17
Remove the description about sass-lint

### LINKED ISSUE #985: sass-lint is unmaintained
Hi there,

The SCSS-Lint's current README [describes sass-lint](https://github.com/sds/scss-lint/blob/8db988731d3cecc4b3c9c63755b364a98858eca9/README.md#notice-consider-other-tools-before-adopting-scss-lint) as an alternative:

> Two alternatives worthy of consideration are stylelint and sass-lint, with sass-lint having closer feature parity to SCSS-lint.

However, sass-lint seems currently unmaintained. (see the sass-lint's [README](https://github.com/sasstools/sass-lint#readme))

So I'm concerned that the description for sass-lint can make users confused (like me). 😓 
I think the description should be removed from the README, but what do you think?

Thanks.

ybiquitous:
Also, this change adds a new description of stylelint plugins in order to make the migration easier.

Fix #985

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,9 +22,9 @@ steps up to help. Since the SCSS-Lint project relies on the Ruby Sass
 implementation, this means it will eventually not support the latest Sass
 features and bug fixes.
 
-Two alternatives worthy of consideration are [stylelint](http://stylelint.io/)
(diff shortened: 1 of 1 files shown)


ybiquitous:
I'm not native in English, so please do not hesitate to give me feedback. 😄

Issue #987, opened 2020-05-04
Request: add scroll-margin property to default whitelist


nali:
The [`scroll-margin` property](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-margin)  has wide support and is in a stable Candidate Recommendation stage: it would be great to add it to the default whitelist.

Happy to open a PR. :)

sds:
Happy to merge a PR, Nali!

nali:
PR #988 now merged. 🥳

PR #988, opened 2020-05-13
Add in scroll-margin properties to default whitelist

nali:
Adds in [scroll-margin](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-margin) to the default whitelist, see #987 .


sds:
Thanks!

PR #992, opened 2021-11-19
Add text-underline-offset as a valid property

anthonyryan1:
Recently started experimenting with this.

MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/text-underline-offset
Caniuse: https://caniuse.com/mdn-css_properties_text-underline-offset


lencioni (approved):
Thanks!

Issue #1000, opened 2023-01-16
Ruby version 3.2.0


elalemanyo:
Hi,
I am trying to upgrade to ruby version 3.2.0, unfortunately there seems to be some error:
```ruby
SCSSLint::Linter::ChainedClasses raised unexpected error linting file app/assets/stylesheets/admin/base/_base.scss: 'undefined method `=~' for a:Sass::Selector::SimpleSequence'
/usr/local/bundle/bundler/gems/scss-lint-7b74cdfb86c0/lib/scss_lint/linter/chained_classes.rb:9:in `block in visit_sequence'
/usr/local/bundle/bundler/gems/scss-lint-7b74cdfb86c0/lib/scss_lint/linter/chained_classes.rb:8:in `each'
/usr/local/bundle/bundler/gems/scss-lint-7b74cdfb86c0/lib/scss_lint/linter/chained_classes.rb:8:in `visit_sequence'
/usr/local/bundle/bundler/gems/scss-lint-7b74cdfb86c0/lib/scss_lint/selector_visitor.rb:12:in `visit_selector_node'
/usr/local/bundle/bundler/gems/scss-lint-7b74cdfb86c0/lib/scss_lint/selector_visitor.rb:5:in `visit_selector'
/usr/local/bundle/bundler/gems/scss-lint-7b74cdfb86c0/lib/scss_lint/selector_visitor.rb:21:in `block in visit_members'
/usr/local/bundle/bundler/gems/scss-lint-7b74cdfb86c0/lib/scss_lint/selector_visitor.rb:20:in `each'
/usr/local/bundle/bundler/gems/scss-lint-7b74cdfb86c0/lib/scss_lint/selector_visitor.rb:20:in `visit_members'
/usr/local/bundle/bundler/gems/scss-lint-7b74cdfb86c0/lib/scss_lint/selector_visitor.rb:14:in `visit_selector_node'
/usr/local/bundle/bundler/gems/scss-lint-7b74cdfb86c0/lib/scss_lint/selector_visitor.rb:5:in `visit_selector'
/usr/local/bundle/bundler/gems/scss-lint-7b74cdfb86c0/lib/scss_lint/linter.rb:139:in `visit'
/usr/local/bundle/bundler/gems/scss-lint-7b74cdfb86c0/lib/scss_lint/linter.rb:154:in `block in visit_children'
[...]

PR #1002, opened 2023-01-21
Add ruby 3.2.0 support

elalemanyo:
#1000

This solves the problems with the tests, but I don't know if it breaks anything on the other side.

BTW: I add a very simple setup to run tests using github actions, I think travis is no longer free but github actions is free for public repositories. You can check it [here](https://github.com/elalemanyo/sds--scss-lint/actions/runs/3975813803)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,26 @@
+name: CI
+
+on:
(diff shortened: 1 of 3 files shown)


sds:
Thank you!

